Issue #366, opened 2026-04-22
refactor: capability-based features for bitrouter binary and bitrouter-api SDK


FrozenArcher404:
## Summary

Refactor the workspace so that **feature flags reflect distinct dependency trees**, and so that the **`bitrouter-api` SDK is the central extension surface**, with sub-crates introduced only when they bring new heavyweight integration dependencies.

A feature exists **only if enabling it pulls a meaningfully different dep tree**. If two flags drag the same crates, they collapse into one. If a flag is purely a module-visibility toggle with no dep-tree delta, it isn't a feature — the code is always built.

In `bitrouter`, features describe **bundles of capability** (the "kind of bitrouter" you want), not individual toggles. Accounts (DB-backed identity / spend) and Payments (MPP + signing) are **core to the product** and baked in unconditionally; they are not behind features.

This issue is a **design proposal**, not an implementation plan.

### 1. `bitrouter` binary features are backend-shaped, and most heavy deps are unconditional anyway

Today's `bitrouter/Cargo.toml` exposes `tui`, `sqlite`, `postgres`, `mysql`, `mpp-tempo`, `mpp-solana`, `mcp`, `rest`. These describe *backends*, not *what the user wants the binary to do*. Several heavy dependencies are already unconditional (`sea-orm`, `ows-*`, `alloy`, `warp`, `reqwest`, `tracing-subscriber`, `dialoguer`, `clap` full, `dirs`) — so the existing flags don't actually shrink the binary.

We embrace that for accounts and payments (they're product-core), but the *interactive operator* and *TUI* surfaces are large, optional dep trees that should be feature-gated.

### 2. `bitrouter-api` (the SDK) under-gates its modules — and over-gates others

`bitrouter-api/src/router/` contains `admin`, `agents`, `agentskills`, `tools`, `models`, `routes`, `context` — compiled unconditionally. Meanwhile `openai`, `anthropic`, `google`, `mcp` are gated as features but **bring no extra dependencies** (pure module gates), so they fail the rule too — they shouldn't be features at all.

There is also no `accounts`, `observe`, or `guardrails` feature on the SDK itself; those integrations live only inside the `bitrouter` binary's `runtime/server.rs`, so external embedders can't pull them through one SDK import.

### A. Crate-creation rule (codified)

[...]

PR #389, opened 2026-04-23
feat(agents): ACP registry install manager + TUI slash commands

SPIKESPIGEL404:
## Summary

Adds a cc-switch-flavoured agent manager to BitRouter, driven by the official [ACP registry](https://cdn.agentclientprotocol.com/registry/v1/latest/registry.json) and surfaced through both the operator CLI and the TUI input bar. The goal is parity with [cc-switch](https://github.com/farion1231/cc-switch)'s provider-management UX, adapted to a terminal interface and to BitRouter's \"default provider is BitRouter\" model (so most of cc-switch's config-rewriter machinery falls away).

Four commits, ~2,000 LOC new, no new crate (per #366). Everything lands in existing crates because nothing pulls a new heavyweight dep.

### Registry-driven agent catalog

- `bitrouter-config::acp` — pure serde types mirroring the registry JSON shape (`RegistryIndex`, `RegistryAgent`, `RegistryDistribution`, `RegistryNpx`, `RegistryUvx`) plus a `registry_agent_to_config` conversion into the runtime `AgentConfig`. Zero async/HTTP deps stay out of this crate.
- `bitrouter-providers::acp::registry` — async fetch with on-disk TTL cache at `<home>/cache/acp-registry.json`. Atomic write (temp + rename) so a crash mid-write leaves the previous cache intact. Offline behaviour: stale cache is served on network failure; built-in agent defs remain the ultimate fallback. `BITROUTER_ACP_REGISTRY_URL` env + `acp_registry_url` field in `bitrouter.yaml` let users pin a private fork or mirror.
- `merge_registry_into_agents` is additive: it fills in ids not already defined by built-ins or the user's YAML, so user config always wins.

### Eager installer

- `bitrouter-providers::acp::eager::install_agent` dispatches to the first viable method:
  - `npx` → `npm install -g <package>` (checks `npx` on PATH, since that's what launch invokes)
  - `uvx` → `uv tool install <package>` (checks `uvx` on PATH)
  - `binary` → download + extract via the existing `install_binary_agent` helper, now fixed to install into `<home>/agents/<id>/` instead of `~/.local/bin` and to record the resolved path.
- `install_binary_agent` signature now takes `install_dir: &Path` explicitly (breaking change for internal callers; the one call site in `bitrouter-tui` is updated).
- Progress streams through the existing `InstallProgress` variants. A single `spawn_progress_forwarder` helper in `bitrouter-tui::app::slash` is shared between the slash-install path and `agent_lifecycle::start_binary_install`, so percent computation stays consistent.

### Install-state ledger

- `bitrouter-providers::acp::state` — JSON array at `<home>/agents/state.json` recording what was installed and how. `InstallMethod` has a `Display` impl used by both the CLI and TUI list views.
- `overlay_install_state_sync` rewrites `AgentConfig.binary` to the persisted resolved path for binary-installed agents, so cold restarts find the installed agent without re-discovery.
- Missing file → empty vec. Corrupt JSON → empty vec + `tracing::warn!`, so install/uninstall operations always recover cleanly.

### `RuntimePaths` extension

- Adds `cache_dir`, `agents_dir`, `agent_state_file`, plus an `agent_install_dir(id)` helper.
- `scaffold_home` mkdirs `cache/` and `agents/` alongside the existing `run/` and `logs/`.

### CLI surface (behind the existing `cli`/`tui` feature bundles)

```
bitrouter agents list [--refresh]
bitrouter agents install <id>
bitrouter agents uninstall <id>
bitrouter agents update [<id>]
bitrouter providers list
bitrouter providers use {default|byok}
```

`agents list` prints the merged view: registry version, install-state method label (npx/uvx/binary), and PATH status. `providers use` is a guidance command (\"to switch, exit and run `bitrouter init`\") since the actual re-gen lives in the existing `init` wizard.

### TUI slash commands

Lines starting with `/` are intercepted in the input bar. Unknown `/...` still falls through to the agent-prompt path so users can talk about literal slashes.

| Command | Effect |
| --- | --- |
| `/agents [list] [--refresh]` | registry + install ledger roll-up, rendered into scrollback |
| `/agents install <id>` | eager install with streamed progress |
| `/agents uninstall <id>` | remove install, clear ledger entry |
| `/agents update [<id>]` | re-install one or all installed agents |
| `/providers [list]` | provider summary with key-status |
| `/providers use {default\|byok}` | guidance pointer |
| `/login`, `/logout`, `/whoami` | hints to exit the TUI and use the CLI |
| `/usage`, `/keys`, `/init` | pointer messages (deferred features) |
| `/help`, `/?` | opens the Help modal |

Long-running operations (install/update) run on `tokio::spawn`ed tasks and emit progress through the existing `AppEvent::InstallProgress` / `InstallComplete` / `InstallFailed` variants. One-shot text output uses a new `AppEvent::SystemMessage { text }` so background tasks can write into the active tab's scrollback.

### Startup registry pre-fetch

Before handing `BitrouterConfig` to the TUI, `main.rs` fetches the registry (cache-first, with the same TTL used by slash commands) and merges new entries into `config.agents`. Closes the gap where registry-only agents were invisible until the user ran `/agents list`. Failure falls through quietly with a one-line note to stderr.

## Crate layout decisions (per #366)

This refactor introduces zero new heavyweight deps:

- `reqwest`, `serde_json`, `tokio/fs`, `flate2`, `tar`, `zip` all already present
- `BitrouterConfig` gains one optional string field
- `RuntimePaths` gains three path fields and one helper

Per issue #366's codified rule (\"new `bitrouter-*` crate only if enabling it grows the SDK's dep tree with new heavyweight integration crates\"), no new crate is justified. Work lands in:

- `bitrouter-config` — serde schema types (`acp` module)
- `bitrouter-providers::acp::{registry, state, eager}` — I/O for fetch, persistence, install
- `bitrouter-tui::app::slash` — TUI surface
- `bitrouter/src/cli/{agents, providers}` — operator CLI

One Cargo.toml fix: the `acp` feature in `bitrouter-providers` now explicitly declares `tokio/fs`. Previously it compiled only because every reachable caller transitively enabled `agentskills` (which happens to declare it) — building `bitrouter-tui` in isolation would have broken.

## Behaviour unchanged

- The existing agent-chat loop, event pump, tab system, permission flow, scrollback streaming, @-mention routing, and modal system.
- The onboarding wizard in `bitrouter init`: BitRouter Cloud vs BYOK remains the first-run choice.
- `builtin_agent_defs` stays as the offline fallback and is not deleted.

## Explicitly out of scope (deferred)

The brief consciously cut these from v1; they're mentioned here so reviewers can anchor against the full cc-switch feature set:

- **MCP panel** — MCP runs client-side per CLI, not over HTTP; no existing data layer in BitRouter to surface.
- **Prompts editor** — CLAUDE.md / AGENTS.md / GEMINI.md editing.
- **Skills install** — the `FilesystemSkillRegistry` exists (`bitrouter-providers/agentskills`) but isn't surfaced.
- **Cloud sync** — WebDAV or otherwise.
- **Deep link** (`bitrouter://`) — per-OS registration work.
- **Real provider add/remove flow** in `bitrouter providers` — today it points users at `bitrouter init`.

Three items noted from the post-implementation review as follow-ups rather than v1 fixes:

- **L1**: `slash_agents_update`'s blocking ledger read (`load_state_sync` inside the TUI event loop). Non-blocking in practice; could be async for architectural consistency.
- **L2**: `state::upsert_record` is not atomic across concurrent callers. User-triggered concurrency is bounded by TUI input serialization, but `/agents update` with ≥2 installed agents could theoretically race. Fix is a `Mutex` around load-modify-save.
- **L3**: Further DRY on the install-progress forwarder patterns (partially done — the TUI paths share a helper, but CLI still has its own stdout-facing reporter).

## Testing

- `cargo build -p bitrouter --features tui` — clean
- `cargo build -p bitrouter-providers --no-default-features --features acp` — clean (verifies the `tokio/fs` gate)
- `cargo clippy --all-features -D warnings` — clean across `bitrouter-config`, `bitrouter-providers`, `bitrouter-tui`, `bitrouter`
- `cargo fmt --all --check` — clean
- `cargo test --all-features` — 324 tests pass across the four crates
  - 4 new in `bitrouter-config::acp`
  - 5 in `bitrouter-providers::acp::state`
  - 4 in `bitrouter-providers::acp::registry` (incl. the new `merge_preserves_existing_and_adds_new`)
  - 2 in `bitrouter-providers::acp::eager`
  - 5 in `bitrouter-tui::app::slash` (parser)

**Pre-existing test note**: `bitrouter-config::config::tests::load_with_custom_derived_provider` is environment-sensitive — it fails when `\$OPENAI_BASE_URL` is set in the shell because the `env_prefix` auto-override kicks in. The test fails on `main` too under the same env and is unchanged by this PR.

## Review findings addressed

A three-reviewer pass surfaced eight findings; all fixed in `fix(acp): review fixes` (the last commit):

- **C1**: `handle_install_complete` racing the slash-install state write (could corrupt `state.json` with `version: \"\"`)
- **C2**: `which(\"npm\")` / `which(\"uv\")` at install time vs `find_on_path(\"npx\")` / `(\"uvx\")` at launch — install-time check now probes the launch binary
- **C3**: `slash_agents_list` ignored the `acp_registry_url` YAML field
- **C4**: `tokio/fs` feature not declared on `acp`
- **I1**: `write_cache` non-atomic — now uses temp + rename like `save_state`
- **I2**: Download percent hardcoded to 0 in slash install — now computed via shared helper
- **I3**: `method_label` duplicated in CLI and TUI — replaced with `Display` on `InstallMethod`
- **I4**: `run_list` hid `block_in_place` inside `pub fn` — now `pub async fn`

## Test plan

- [ ] `bitrouter agents list` with network available shows the full registry plus install status
- [ ] `bitrouter agents list --refresh` bypasses the cache
- [ ] `bitrouter agents install claude-acp` (npx-distributed) installs successfully; subsequent `bitrouter agents list` marks it installed
- [ ] `bitrouter agents install codex-acp` (binary-distributed) downloads to `~/.bitrouter/agents/codex-acp/`; resolved path persists to `state.json`
- [ ] Cold restart after binary install: `resolve_launch` picks up the persisted path without re-discovery
- [ ] `bitrouter agents uninstall codex-acp` removes the install dir and clears the ledger entry
- [ ] `bitrouter agents update` iterates all installed agents
- [ ] `bitrouter providers list` prints config providers with key-status markers
- [ ] `BITROUTER_ACP_REGISTRY_URL=<custom>` overrides the registry URL; `acp_registry_url` in `bitrouter.yaml` also works (env wins)
- [ ] Offline start: TUI launches, prints \"ACP registry unavailable\" note, proceeds with built-ins
- [ ] TUI `/agents install claude-acp` streams progress into the active tab; completion message appears with method label
- [ ] TUI `/agents list` honours `acp_registry_url` from YAML (not just env)
- [ ] TUI `/providers list`, `/help`, `/whoami` (hint), `/init` (hint) render as system messages
- [ ] Unknown `/xyz` line falls through and is sent to the active agent as a prompt
- [ ] `/agents update` with ≥1 installed agent runs sequentially per agent
- [ ] cc-switch-style registry-only agents appear in the TUI agent tab list without the user running `/agents list` first (pre-fetch check)

## Depends on

- #388 is a separate auth refactor (cloud device-flow login + host-minted JWT verification) that was pulled out of this branch. These two PRs are independent and can land in either order.


FrozenArcher404:
@copilot+gpt-5.4 Many CI workflows fail to run, including clippy and `bitrouter` feature builds, in this PR. Fix them.

PR #425, opened 2026-05-03
feat(core,api)!: TargetOverlay + per-endpoint key overrides

SPIKESPIGEL404:
## Summary

Adds the primitives needed for per-caller credential injection (e.g. a "bring your own key" overlay in a downstream cloud deployment) without forking the routing pipeline. The change is additive at the type level and the breaking-change marker is for one positional-argument addition on three filter constructors (see below).

## Why

Today, two gaps make per-request credential injection awkward:

1. `Endpoint.api_key` and `Endpoint.api_base` (per-endpoint config overrides) are resolved by `ConfigRoutingTable` into `ResolvedTarget` — and then **dropped** at the `RoutingTable` trait boundary. The public `RoutingTarget` carries only `provider_name`, `service_id`, `api_protocol`, so `LanguageModelRouter::route_model()` never sees them.
2. Neither `RoutingTable::route()` nor `LanguageModelRouter::route_model()` sees the request's `CallerContext`, so a downstream router has no way to look up "the credential for *this caller* on the resolved provider."

Both are solvable inside `bitrouter-core` + `bitrouter-config` + `bitrouter-api` with additive primitives.

## What

- **`bitrouter-core`** — `RoutingTarget` gains `api_key_override: Option<String>` and `api_base_override: Option<String>`. New `TargetOverlay` trait (dynosaur-wrapped as `DynTargetOverlay`) provides the post-route, pre-instantiate hook with `(&mut RoutingTarget, &CallerContext)`. The trait doc-comment specifies the expected `BitrouterError` variants for credential failures so client-visible errors stay actionable (`AccessDenied` -> 403, `InvalidRequest` -> 400).
- **`bitrouter-config`** — `ConfigRoutingTable::route()` and `ConfigToolRoutingTable::route()` now propagate the override fields from `ResolvedTarget` into `RoutingTarget`. Fixes the silent drop.
- **`bitrouter-api`** — `GateContext` carries `Option<Arc<DynTargetOverlay<'static>>>`. The three `*_filter_with_payment_gate` constructors take it as an explicit positional argument; the `*_with_mpp` peers default it to `None` (with an inline comment explaining why). The handlers invoke the overlay between `table.route()` and `router.route_model()`.

## Breaking change

`chat_completions_filter_with_payment_gate`, `messages_filter_with_payment_gate`, and `responses_filter_with_payment_gate` gain a `target_overlay: Option<Arc<DynTargetOverlay<'static>>>` parameter, placed before `account_filter`. Existing callers can pass `None` to retain current behaviour.

`RoutingTarget` field additions are technically struct-API changes, but I checked the repo for destructuring patterns — only construction sites exist, all updated in this PR.

## Test plan

- [x] `cargo fmt -- --check` clean
- [x] `cargo clippy --all-features --all-targets` clean (no warnings)
- [x] `cargo test --all-features` — all tests pass except one pre-existing failure (`bitrouter-config::config::tests::load_with_custom_derived_provider`) which fails on clean `origin/main` due to env-prefix substitution picking up a local `OPENAI_BASE_URL`. Unrelated to this change.
- [x] New unit tests:
  - `bitrouter-core::routers::router::tests::dyn_overlay_mutates_target_through_arc` — exercises the dynosaur wrapper round-trip and `Arc<DynTargetOverlay>` invocation shape that filters use.
  - `bitrouter-config::routing::tests::route_trait_propagates_endpoint_overrides` — locks in the trait-level fix for the silent drop of `Endpoint.api_key`/`api_base`.

### DIFF
--- a/bitrouter-api/src/mpp/gate_context.rs
+++ b/bitrouter-api/src/mpp/gate_context.rs
@@ -3,6 +3,7 @@
 use std::sync::Arc;
 
 use bitrouter_core::observe::{CallerContext, MetadataHook, ObserveCallback};
+use bitrouter_core::routers::router::DynTargetOverlay;
 
 use super::PaymentGate;
 
@@ -18,4 +19,8 @@ pub struct GateContext {
     pub observer: Arc<dyn ObserveCallback>,
     pub metadata_hook: MetadataHook,
     pub origin: Option<String>,
+    /// Optional per-request hook that mutates the routing target after
+    /// resolution. Invoked between `RoutingTable::route()` and
+    /// `LanguageModelRouter::route_model()`. `None` is the no-op default.
+    pub target_overlay: Option<Arc<DynTargetOverlay<'static>>>,
 }
--- a/bitrouter-api/src/router/admin.rs
+++ b/bitrouter-api/src/router/admin.rs
@@ -215,6 +215,8 @@ mod tests {
                     provider_name: "openai".to_owned(),
                     service_id: "gpt-4o".to_owned(),
                     api_protocol: ApiProtocol::Openai,
+                    api_key_override: None,
+                    api_base_override: None,
                 })
             } else {
                 Err(BitrouterError::invalid_request(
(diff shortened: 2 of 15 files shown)


SPIKESPIGEL404:
@claude fix issues that caused CI build fail

SPIKESPIGEL404:
@claude fix fmt error

PR #453, opened 2026-05-11
fix: propagate reasoning content across all streaming protocols

### LINKED ISSUE #448: Reasoning/thinking content silently dropped in all streaming protocol paths
## Summary

Thinking/reasoning content is silently dropped in **every protocol path** (OpenAI Chat, OpenAI Responses, Google, and Anthropic output-side converters). This causes an apparent silence of up to ~1 minute before the first visible stream chunk arrives, as reported by users routing `v1/messages` through OpenAI-compatible providers (e.g. GLM-5.1 via Aliyun) that support extended thinking.

## Root Cause

There are two systematic gaps across all protocols:

1. **Input side**: Each provider's SSE parser/types don't extract reasoning content from provider-specific streaming formats — it's silently discarded by serde or ignored by the stream state machine.
2. **Output side**: Every `StreamConverter.convert()` drops `ReasoningStart/Delta/End` `LanguageModelStreamPart` variants via catch-all patterns (`_ => vec![]` or `_ => None`).

Even Anthropic-to-Anthropic streaming drops thinking on the output side, despite the upstream provider correctly parsing it.

Additionally, all provider request converters **error** on `LanguageModelAssistantContent::Reasoning` in assistant prompts, breaking multi-turn conversations with thinking-enabled models.

---

### 1. OpenAI Chat Completions (the reported GLM-5.1/Aliyun path)

**Input side:**
- `ChatCompletionChunkDelta` missing `reasoning_content` field — serde silently discards it from SSE chunks.  
  📍 `bitrouter-core/src/api/openai/chat/types.rs:261-268`
- `OpenAiStreamState::apply_chunk` never emits `ReasoningStart/Delta/End` — only processes `delta.content` and `delta.tool_calls`.  
  📍 `bitrouter-providers/src/openai/chat/api.rs:853-901`
- Assistant prompt with `Reasoning` content block returns an error instead of stripping it.  
  📍 `bitrouter-providers/src/openai/chat/api.rs:474-479`

**Output side:**
- Catch-all `_ => None` drops reasoning parts.  
  📍 `bitrouter-core/src/api/openai/chat/convert.rs:245,335`

### 2. OpenAI Responses API

**Input side:**
- SSE parser ignores `response.reasoning_text.delta` events — the test at `api.rs:1926` explicitly confirms they produce empty parts and are silently discarded.  
  📍 `bitrouter-providers/src/openai/responses/api.rs:1926-1948`
- No reasoning fields in Responses delta types.  
  📍 `bitrouter-core/src/api/openai/responses/types.rs`
- Assistant prompt with `Reasoning` content block returns an error.  
  📍 `bitrouter-providers/src/openai/responses/api.rs:492-497`

**Output side:**
- Catch-all `_ => vec![]` drops reasoning parts.  
  📍 `bitrouter-core/src/api/openai/responses/convert.rs:285`

### 3. Google Generative AI

**Input side:**
- No reasoning/thought fields in Google types (`GooglePart`, `GoogleContent`, etc.).  
  📍 `bitrouter-core/src/api/google/generate_content/types.rs`
- Google SSE parser has no handling for `thoughtContent` / reasoning parts in streaming chunks.  
  📍 `bitrouter-providers/src/google/generate_content/api.rs`
- Assistant prompt with `Reasoning` content block returns an error.  
  📍 `bitrouter-providers/src/google/generate_content/api.rs:559-568`

**Output side:**
- Catch-all `_ => None` drops reasoning parts.  
  📍 `bitrouter-core/src/api/google/generate_content/convert.rs:284`

### 4. Anthropic Messages (Anthropic-to-Anthropic path)

**Input side:**
- ✅ Works correctly — the provider parses `Thinking` blocks and emits `ReasoningStart/Delta/End` stream parts.  
  📍 `bitrouter-providers/src/anthropic/messages/api.rs:835-975`
- ❌ Assistant prompt with `Reasoning` content block returns an error (breaks multi-turn).  
  📍 `bitrouter-providers/src/anthropic/messages/api.rs:514-522`

**Output side:**
- Catch-all `_ => vec![]` drops reasoning parts — even Anthropic-to-Anthropic streaming drops thinking on the output side.  
  📍 `bitrouter-core/src/api/anthropic/messages/convert.rs:289`

---

## Impact

- **User-visible delay**: When routing through thinking-enabled OpenAI-compatible providers, the client sees no SSE events until the first `content` chunk arrives after thinking finishes (~1 minute of apparent silence). The connection stays alive (keep-alive pings at 15s intervals via `sse.rs`), so the client doesn't disconnect, but nothing is rendered.
- **Token accounting**: Reasoning tokens are partially tracked (e.g. `reasoning_tokens` in OpenAI usage, `output_tokens.reasoning` in core usage), but the actual reasoning content is never delivered to the client.
- **Multi-turn broken**: All providers error on reasoning content in assistant prompts, making multi-turn conversations with thinking models impossible (prior turn thinking blocks can't be echoed back).

---

## Fix Scope

The fix requires changes in 4 areas:

1. **Provider SSE delta types** — Add `reasoning_content` to `ChatCompletionChunkDelta`, reasoning event types to Responses parser, `thoughtContent`/`thought` to Google types, and ensure Anthropic types cover all thinking variants.

2. **Provider stream state machines** — Each provider's SSE state machine (`OpenAiStreamState`, `OpenAiResponsesSseParser`, `GoogleSseParser`, Anthropic SSE parser) must emit `ReasoningStart/Delta/End` `LanguageModelStreamPart` variants when reasoning content is encountered.

3. **Output-side StreamConverters** — All 4 converters (Anthropic `convert.rs:289`, OpenAI Chat `convert.rs:245,335`, OpenAI Responses `convert.rs:285`, Google `convert.rs:284`) must handle `ReasoningStart/Delta/End` and convert them to protocol-specific thinking/reasoning SSE events:
   - Anthropic → `thinking` content block start/delta/stop events
   - OpenAI Chat → `reasoning_content` in delta
   - OpenAI Responses → `response.reasoning_text.delta` events
   - Google → `thoughtContent` parts

4. **Assistant prompt handling** — All 4 providers' `convert_prompt`/request builders must **strip** reasoning content from assistant prompts instead of returning an error, so multi-turn conversations work.

---

## Reported By

User routing GLM-5.1 through: `v1/messages → bitrouter → upstream Aliyun coding plan OpenAI-compatible endpoint`. First visible stream part takes ~1 minute to arrive, but the streaming connection is not lost and everything works fine once text content starts.

FrozenArcher404:
## Summary

Closes #448.

Thinking/reasoning content was silently dropped in **every** protocol path (OpenAI Chat, OpenAI Responses, Google, and Anthropic output-side converters), causing ~1 minute of apparent SSE silence when routing thinking-enabled upstreams (e.g. GLM-5.1 via Aliyun) through bitrouter's `v1/messages` endpoint.

The fix closes the gaps systematically in three areas:

- **Input** (provider SSE parsers): extract `reasoning_content` from OpenAI-compatible chunks, `response.reasoning_text.delta`/`.done` from Responses, and `thought: true` parts from Gemini 2.5+; emit `ReasoningStart/Delta/End` core stream parts. Reasoning blocks close cleanly when text starts and are flushed on finish.
- **Output** (`StreamConverter` for each protocol): map `ReasoningStart/Delta/End` to the protocol-specific carrier — `reasoning_content` (OpenAI Chat), `response.reasoning_text.delta`/`.done` (OpenAI Responses), `GooglePart { thought: true }` (Gemini), and Anthropic `thinking` content_block_start / `thinking_delta` / content_block_stop events. Thinking blocks auto-close on text/tool transitions and at finish.
- **Assistant prompt**: silently strip `LanguageModelAssistantContent::Reasoning` across all four providers instead of erroring, so multi-turn conversations with thinking models work.

External integration references (DeepSeek/OpenAI/Gemini/Anthropic doc URLs) are inlined as comments around each conversion path.

## Test plan

- [x] `cargo nextest run --all-features` — 899 tests pass (+16 new)
- [x] `cargo clippy --all-features` — clean
- [ ] Manual smoke test: `v1/messages` → bitrouter → OpenAI-compatible thinking provider; verify SSE thinking events arrive in real time
- [ ] Manual smoke test: `v1/chat/completions` with DeepSeek-R1 or GLM; verify `reasoning_content` deltas surface
- [ ] Manual smoke test: Gemini 2.5+ with thinking enabled; verify `thought: true` parts surface
- [ ] Manual smoke test: Anthropic extended thinking round-trip; verify `thinking_delta` arrives before `text_delta`

New unit tests cover both input (SSE state machine) and output (`StreamConverter`) for each of the four protocols, plus boundary conditions (text-after-reasoning closure, reasoning-only stream finish flush, multi-turn echo-back).

### DIFF
--- a/bitrouter-api/src/router/openai/responses/filters.rs
+++ b/bitrouter-api/src/router/openai/responses/filters.rs
@@ -118,7 +118,7 @@ where
     let options = convert::to_call_options(request);
 
     if is_stream {
-        handle_stream(&model, options).await
+        handle_stream(&model, options, model_id).await
     } else {
         let result = model
             .generate(options)
@@ -366,14 +366,22 @@ where
             request_id: Some(request_id.clone()),
         };
 
+        let stream_model_id = model_id.clone();
         tokio::spawn(async move {
             let mut stream = stream_result.stream;
-            let mut converter = convert::StreamConverter::new();
+            let mut converter = convert::StreamConverter::new(stream_model_id);
             use tokio_stream::StreamExt as _;
             let mut observation = crate::router::StreamObservation::new();
             let mut client_disconnected = false;
+            let mut saw_finish = false;
             while let Some(part) = stream.next().await {
                 observation.record_part(&part);
+                if matches!(
+                    part,
+                    bitrouter_core::models::language::stream_part::LanguageModelStreamPart::Finish { .. }
+                ) {
+                    saw_finish = true;
+                }
                 let events = converter.convert(&part);
                 if !events.is_empty() && !metered.deduct_or_pause(&tx).await {
                     client_disconnected = true;
@@ -395,9 +403,22 @@ where
                     break;
                 }
             }
-            let _ = tx
-                .send(Ok(warp::sse::Event::default().data("[DONE]")))
-                .await;
+            // Ensure the stream closes with `response.completed` per the
+            // Responses spec; Codex CLI retries otherwise.
+            // https://platform.openai.com/docs/api-reference/responses-streaming
+            if !saw_finish && !client_disconnected {
+                for event in converter.finish() {
+                    let data = serde_json::to_string(&event).unwrap_or_default();
+                    let sse = Ok(warp::sse::Event::default()
+                        .event(&event.event_type)
+                        .data(data));
+                    if tx.send(sse).await.is_err() {
+                        client_disconnected = true;
+                        break;
+                    }
+                }
+            }
+            // Responses SSE has no [DONE] terminator.
 
             let latency_ms = start.elapsed().as_millis() as u64;
             let ctx = RequestContext {
@@ -568,6 +589,7 @@ where
         handle_stream_with_observe(
             &model,
             options,
+            model_id,
             crate::router::StreamObserveContext {
                 observer,
                 route: incoming_model,
@@ -628,6 +650,7 @@ where
 async fn handle_stream(
     model: &(impl LanguageModel + ?Sized),
     options: bitrouter_core::models::language::call_options::LanguageModelCallOptions,
+    model_id: String,
 ) -> Result<Box<dyn warp::Reply>, warp::Rejection> {
     let stream_result = model
         .stream(options)
@@ -639,9 +662,16 @@ async fn handle_stream(
 
     tokio::spawn(async move {
         let mut stream = stream_result.stream;
-        let mut converter = convert::StreamConverter::new();
+        let mut converter = convert::StreamConverter::new(model_id);
         use tokio_stream::StreamExt as _;
+        let mut saw_finish = false;
         'outer: while let Some(part) = stream.next().await {
+            if matches!(
+                part,
+                bitrouter_core::models::language::stream_part::LanguageModelStreamPart::Finish { .. }
+            ) {
+                saw_finish = true;
+            }
             for event in converter.convert(&part) {
                 let data = serde_json::to_string(&event).unwrap_or_default();
                 let sse = Ok(warp::sse::Event::default()
@@ -652,9 +682,24 @@ async fn handle_stream(
                 }
             }
         }
-        let _ = tx
-            .send(Ok(warp::sse::Event::default().data("[DONE]")))
-            .await;
+        // Upstream closed without a Finish part — synthesize a final
+        // `response.completed` so strict clients (e.g. Codex CLI) don't
+        // hit "stream closed before response.completed" and retry.
+        // https://platform.openai.com/docs/api-reference/responses-streaming
+        if !saw_finish {
+            for event in converter.finish() {
+                let data = serde_json::to_string(&event).unwrap_or_default();
+                let sse = Ok(warp::sse::Event::default()
+                    .event(&event.event_type)
+                    .data(data));
+                if tx.send(sse).await.is_err() {
+                    break;
+                }
+            }
+        }
+        // Responses SSE has no [DONE] terminator; `response.completed` is
+        // the canonical close. Emitting [DONE] here would cause Codex to
+        // surface a parse error and retry.
     });
 
     let sse_stream = tokio_stream::wrappers::ReceiverStream::new(rx);
@@ -664,6 +709,7 @@ async fn handle_stream(
 async fn handle_stream_with_observe(
     model: &(impl LanguageModel + ?Sized),
     options: bitrouter_core::models::language::call_options::LanguageModelCallOptions,
+    model_id: String,
     ctx: crate::router::StreamObserveContext,
 ) -> Result<Box<dyn warp::Reply>, warp::Rejection> {
     let stream_result = model
@@ -688,12 +734,19 @@ async fn handle_stream_with_observe(
 
     tokio::spawn(async move {
         let mut stream = stream_result.stream;
-        let mut converter = convert::StreamConverter::new();
+        let mut converter = convert::StreamConverter::new(model_id);
         use tokio_stream::StreamExt as _;
         let mut observation = crate::router::StreamObservation::new();
         let mut client_disconnected = false;
+        let mut saw_finish = false;
         while let Some(part) = stream.next().await {
             observation.record_part(&part);
+            if matches!(
+                part,
+                bitrouter_core::models::language::stream_part::LanguageModelStreamPart::Finish { .. }
+            ) {
+                saw_finish = true;
+            }
             let mut send_failed = false;
             for event in converter.convert(&part) {
                 let data = serde_json::to_string(&event).unwrap_or_default();
@@ -710,9 +763,23 @@ async fn handle_stream_with_observe(
                 break;
             }
         }
-        let _ = tx
-            .send(Ok(warp::sse::Event::default().data("[DONE]")))
-            .await;
+        // Synthesize `response.completed` when the upstream stopped without
+        // a Finish so Codex CLI sees a well-formed terminator instead of
+        // raising "stream closed before response.completed" and retrying.
+        // https://platform.openai.com/docs/api-reference/responses-streaming
+        if !saw_finish && !client_disconnected {
+            for event in converter.finish() {
+                let data = serde_json::to_string(&event).unwrap_or_default();
+                let sse = Ok(warp::sse::Event::default()
+                    .event(&event.event_type)
+                    .data(data));
+                if tx.send(sse).await.is_err() {
+                    client_disconnected = true;
+                    break;
+                }
+            }
+        }
+        // Responses SSE has no [DONE] terminator.
 
         let latency_ms = start.elapsed().as_millis() as u64;
         let ctx = RequestContext {
--- a/bitrouter-api/src/router/openai/responses/tests.rs
+++ b/bitrouter-api/src/router/openai/responses/tests.rs
@@ -339,22 +339,59 @@ async fn responses_streaming_sends_sse_events() {
     assert_eq!(res.headers()["content-type"], "text/event-stream");
 
     let events = parse_sse_body(res.body());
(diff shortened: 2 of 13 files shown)


FrozenArcher404:
## Follow-up: Responses streaming lifecycle for Codex

Pushed a second commit that fixes the related issue where Codex CLI repeatedly retries against `/v1/responses` with `stream closed before response.completed`.

Root causes:
- The `StreamConverter` emitted bare `response.output_text.delta` / `response.function_call_arguments.delta` events without the required `response.created` / `response.in_progress` / `response.output_item.added` / `response.content_part.added` preamble and matching `*.done` closures, and without `sequence_number`.
- `response.completed` was emitted with an empty payload — Codex needs the full `response` object with `status` and `usage`.
- The handler appended a `[DONE]` SSE line, which the Responses protocol doesn't use (Codex sees it as malformed and retries).

Reference: <https://platform.openai.com/docs/api-reference/responses-streaming>

What changed:
- `ResponsesStreamEvent`: added `sequence_number` (required), `response`, `text`, `part` fields.
- `StreamConverter` rewritten as a stateful lifecycle machine that lazily opens message / reasoning / function_call output items, emits the full canonical sequence (created → in_progress → output_item.added → content_part.added → deltas → *.done → output_item.done → completed), assigns monotonic `sequence_number`, and adopts the upstream response id from `ResponseMetadata` before `response.created` so `previous_response_id` chaining works.
- Constructor takes `model_id`. New `finish()` method synthesizes `response.completed` if the upstream closes without a Finish part.
- All three Responses stream handlers (no-observe, observe, MPP gated) drop `[DONE]` and call `finish()` on premature close.
- Tests updated for the new shape; new tests cover sequence_number monotonicity, no-`[DONE]` invariant, full `response.completed` envelope, upstream response-id adoption, and `finish()` synthesis.

All 903 tests pass; clippy + fmt clean.

FrozenArcher404:
## Follow-up: Codex CLI multi-turn request compatibility

Pushed commit `f7e450b` fixing the 400 error when Codex CLI sends a multi-turn `/v1/responses` request to bitrouter:

> `failed to deserialize request body into ResponsesRequest: data did not match any variant of untagged enum ResponsesInput at line 1 column 40262`

Inspected the actual Codex session rollout and the codex-rs/protocol/src/models.rs `ResponseItem` enum and found three shapes our deserializer rejected:

1. **Unknown input item types** — Codex's `input` carries `reasoning`, `web_search_call`, `local_shell_call`, `image_generation_call`, `custom_tool_call`, `compaction`, etc. after the first turn. The untagged enum had no fall-through, so a single unknown item failed the whole `ResponsesInput`. Added an `Unknown(serde_json::Value)` catch-all; `convert_input_items` silently drops these since bitrouter can't replay them to a non-Responses upstream.
2. **Top-level `instructions` field** — Codex sends the developer prompt here rather than as a system message. The field wasn't declared, so serde dropped it silently and the routed model saw no system context. Added `instructions: Option<String>` to `ResponsesRequest` and lift it into a leading `System` message in `to_call_options`.
3. **Structured `function_call_output.output`** — Codex emits multimodal tool results as `{content_items: [{type: "input_text", text: "..."}]}`. The field was typed `String` and 400'd on the structured form. Changed to `serde_json::Value` and added a `stringify_function_call_output` helper that collapses `content`/`content_items` into newline-joined text.

Reference: <https://platform.openai.com/docs/api-reference/responses/create>

Tests added: realistic Codex multi-turn payload (instructions + developer/user/assistant messages + reasoning + web_search_call + local_shell_call), instructions lifting, unknown-item skipping, both string and structured `function_call_output` shapes. 907 tests pass; clippy and fmt clean.

FrozenArcher404:
## Follow-up: Codex → chat-completions upstream payload shape

Pushed commit `1fe89fa` fixing the Z.ai/GLM "prompt parameter was not received normally" 400 reported when routing Codex CLI (Responses) through opencode-go's chat-completions GLM-5.1.

Found the root cause by inspecting the rollout log: every Codex turn emits two `response_item`s — a `message` (role=`assistant`, text=`""`) and a separate `reasoning` (the real content). Two bugs in `convert_input_items` compounded:

1. **role=assistant was demoted to User.** The match arm caught `"system"|"developer"` and dumped everything else into `User`. After many turns the prompt was a wall of consecutive user messages with no assistant turns — upstreams that enforce alternation reject this. Added an explicit `"assistant"` arm producing `Assistant` with `Text` content; empty content is filtered so vacuous shells don't survive.

2. **Reasoning items dropped silently.** Combined with the empty-text assistant messages, the routed model saw zero assistant context. The `Unknown` catch-all now sniffs for `type:"reasoning"` and lifts the inner `content[].text` into an `Assistant` message with `Reasoning` content. Providers that round-trip thinking (Anthropic) get it for free; chat-completions providers strip it on the way out.

Defense in depth in `bitrouter-providers/src/openai/chat/api.rs`: `convert_prompt` now skips assistant messages that, after stripping `Reasoning`, have no `Text` and no `tool_calls` — so any path that produces an empty assistant turn gets a clean upstream payload, not just Codex.

Tests added:
- `to_call_options_maps_assistant_role_to_assistant_message`
- `to_call_options_drops_empty_assistant_messages_from_codex`
- `to_call_options_lifts_reasoning_items_to_assistant_reasoning`
- `to_call_options_drops_other_unknown_items` (web_search_call etc. still dropped)
- `convert_prompt_drops_reasoning_only_assistant_turn` (chat-completions side)
- `convert_prompt_keeps_assistant_with_text`

912 tests pass; `cargo clippy --all-features --tests -- -D warnings` clean; `cargo fmt --check` clean.

FrozenArcher404:
## Follow-up: response.completed usage payload hangs Codex TUI

Pushed commit `469ae69` fixing the "Codex finishes turn 1 then hangs forever, no second request" symptom.

Read Codex's actual SSE parser at [codex-rs/codex-api/src/sse/responses.rs](https://github.com/openai/codex/blob/main/codex-rs/codex-api/src/sse/responses.rs) and found:

```rust
struct ResponseCompletedUsage {
    input_tokens: i64,                                      // REQUIRED
    input_tokens_details: Option<ResponseCompletedInputTokensDetails>,
    output_tokens: i64,                                     // REQUIRED
    output_tokens_details: Option<ResponseCompletedOutputTokensDetails>,
    total_tokens: i64,                                      // REQUIRED
}
```

The three core token counts are non-optional `i64`. Bitrouter's `ResponsesUsage` typed them as `Option<u32>` without `skip_serializing_if`, so when the routed upstream returned a Finish part with any count missing (`None`) we serialized that field as JSON `null` inside the `usage` object. Codex's `serde_json::from_value::<ResponseCompleted>` then failed with `invalid type: null, expected i64`. Worse, that parse error is stored in `response_error` and only surfaced when the SSE stream EOFs — by which point the oneshot `LastResponse` channel was never signaled, so the TUI sat forever waiting for the turn to finalize. No retry, no second request.

Fix in `StreamConverter::build_completed_response`:
- If both upstream input *and* output counts are `None`, omit the `usage` object entirely (Codex's `usage: Option<...>` deserializes the missing key as `None` cleanly).
- Otherwise default any missing field to `0` and always emit the three core counts together — so `usage` is never partially-null on the wire.

Defense in depth: added `skip_serializing_if = "Option::is_none"` on the `ResponsesUsage` base fields so any future code path constructing a partial `ResponsesUsage` doesn't reintroduce null token counts.

New regression tests:
- `stream_converter_completed_payload_parses_as_codex_response_completed` — constructs a struct shaped exactly like Codex's `ResponseCompletedUsage` (with non-optional `i64`) and asserts the response.completed JSON parses into it.
- `stream_converter_completed_omits_usage_when_upstream_has_none` — verifies the `usage` key is absent (not null) when no token counts are known.

914 tests pass; clippy + fmt clean.

PR #464, opened 2026-05-18
refactor(cli)!: split bitrouter into lib + cli, merge tui

SPIKESPIGEL404:
## Summary

Restructure the workspace around a **headless-first** product stance. The CLI becomes the canonical operator surface; the TUI becomes an explicit opt-in.

- `bitrouter/` is now library-only — the runtime (HTTP server, router, payment middleware, daemon manager, paths, OAuth token store, migrations).
- `bitrouter-cli/` is a new crate that produces the `bitrouter` binary — every subcommand, the `init` wizard, OAuth login flows, and the merged-in TUI module (gated by the `tui` feature).
- `bitrouter-tui/` is deleted; its source lives at `bitrouter-cli/src/tui/`.

Supersedes [`specs/cli-tui-ops-layer.md`](../blob/refactor/cli-runtime-split/specs/cli-tui-ops-layer.md), which was based on treating CLI and TUI as co-equal product surfaces.

## Design pillars

1. **Headless-first** — bare `bitrouter` starts the proxy in the foreground when configured; runs the `init` wizard otherwise. Scriptable and CI-safe by default.
2. **TUI-as-optional** — the terminal UI never auto-launches. Opt in with `bitrouter tui` (requires the build to have `--features tui`).
3. **CLI-onboarding** — every BitRouter operation flows through CLI subcommands. Lays the groundwork for the cloud-onboarding work in phase 2 (CLI as an autonomous-agent-friendly control plane against `bitrouter-cloud`).

## UX changes (breaking)

| Invocation | Before | After |
|---|---|---|
| `bitrouter` (unconfigured) | init wizard → auto-launches TUI | init wizard → exits with hint to re-run |
| `bitrouter` (configured) | prints status/help | starts the proxy in the foreground (= `bitrouter serve`) |
| `bitrouter tui` | — (didn't exist) | launches the multi-agent ACP UI (requires `--features tui`) |
| `--no-tui` flag | suppresses auto-launch | **removed** (no longer needed; TUI is always opt-in) |
| `bitrouter agents`, `bitrouter agent-proxy` | TUI-feature-gated | always available |

## Feature-flag changes

- `bitrouter` (runtime): dropped `cli` and `tui`. Kept DB driver features (`sqlite`/`postgres`/`mysql`) and payment-chain features (`tempo`/`solana`). ACP support is always-on in the runtime (see TODO §1).
- `bitrouter-cli`: new `tui` feature gates the terminal UI module and the `bitrouter tui` subcommand. DB/chain/protocol features pass through to `bitrouter`.

## Dependency cleanup

- Dropped dead deps from `bitrouter/Cargo.toml`: `tracing-subscriber`, `uuid`, `base64`, `rand`, and the `blocking` feature of `reqwest`.
- Dropped unused direct deps from `bitrouter-cli/Cargo.toml`: `bitrouter-api`, `bitrouter-observe` (still pulled transitively through `bitrouter`).

## Quality bar

- `cargo check --workspace --all-features` — clean (one pre-existing dead-field warning on `TuiConfig` reserved for future status-bar wiring per `specs/bitrouter-tui-product.md`).
- `cargo clippy --workspace --all-features` — clean.
- `cargo fmt --check` — clean.
- `cargo test --workspace --all-features` — all pass except `bitrouter-config::config::tests::load_with_custom_derived_provider` (pre-existing, env-var pollution unrelated to this change; verified by reproducing on `main`).
- Review: three code-reviewer agents ran on the change. Findings 1–5 from their reports are addressed in this PR; findings 6–9 are documented as follow-ups below.

## Future TODOs (intentionally out of scope)

**1. Gate ACP behind a feature on the runtime crate.**
Today `bitrouter-providers/acp` is always-on in `bitrouter` (so `ConfigAgentRouter`, `agent_session_cleanup_loop`, and the `/agents` route are always compiled). This forces pure-proxy embedders to pay the `agent-client-protocol` + `flate2` + `tar` + `zip` compile cost even when they have no agents. The right fix is an `acp` feature on `bitrouter` (default-on) that gates this code, mirrored by an `acp` pass-through feature on `bitrouter-cli`. The `tui` feature would imply `acp`. Architecturally clean but touches enough files (router.rs, server.rs, the CLI agent subcommands, both Cargo.tomls) to deserve its own PR.

**2. Fix the `std::env::set_var` UB in `ensure_ows_passphrase`.**
`bitrouter-cli/src/main.rs:1021` mutates the process environment after `tokio::spawn` has already started the `update_check` task on a worker thread. This is UB on most platforms (Rust 1.81 turned it into a deny-by-default lint). Pre-existing — predates this reorg — but a separate fix should either collect the passphrase before spawning, or thread it through a `OnceLock` so `ows-lib` can read it without process-env mutation.

**3. Phase 2 — CLI-onboarding against `bitrouter-cloud`.**
The cloud control plane already exists at `https://bitrouter.ai` and the `login`/`logout`/`whoami` subcommands use RFC 8628 device-code auth against it. Phase 2 will extend the CLI so an autonomous agent can perform all bitrouter operations against the cloud without a local daemon — e.g. `bitrouter route add foo openai:gpt-4o` against either the local daemon (if running) or the cloud (if logged in). Needs a cloud-API client crate and an auto-detect-or-explicit `--target local|cloud` flag scheme.

**4. Tighten `RuntimePaths` public surface.**
`pub use error::RuntimeError` from `bitrouter::runtime::*` would shorten CLI call sites (two places in `main.rs` use the full `bitrouter::runtime::error::RuntimeError` path). Tiny convenience; defer until something else is in that file.

**5. Headless-CLI consistency pass.**
Items from the superseded `specs/cli-tui-ops-layer.md` that are still relevant to headless-first: `--output text|json` everywhere, stdout/stderr discipline (decorative banners on stderr, data rows on stdout), verb consistency (`route rm` → `route delete`). These are independent of the crate split and deserve their own PR(s).

## Test plan

- [ ] `cargo build --release -p bitrouter-cli` produces a working `bitrouter` binary (headless build).
- [ ] `cargo build --release -p bitrouter-cli --features tui` produces a binary where `bitrouter tui` works.
- [ ] `bitrouter` (no args, no config) launches the `init` wizard and exits cleanly when the wizard completes.
- [ ] `bitrouter` (no args, configured) starts the proxy on `127.0.0.1:8787`.
- [ ] `bitrouter serve` continues to work identically.
- [ ] `bitrouter start` / `stop` / `restart` / `reload` still control the background daemon via the PID file.
- [ ] `bitrouter agents list` works without `--features tui` (previously TUI-gated).
- [ ] `bitrouter --no-tui` returns a clear "unknown argument" error (flag removed).
- [ ] `bitrouter tui` is absent from `--help` when built without `--features tui`.
- [ ] `bitrouter --headless` continues to return the existing deprecation error.


SPIKESPIGEL404:
## Update — Addresses follow-up TODO #1 + dep-graph cleanup

Two new commits extend this PR with the dep-cleanup follow-up work, paving the way for the headless ACP CLI (separate future PR).

### What landed

- **`72efbab4` — `feat(bitrouter): gate ACP behind feature; add providers/accounts re-export modules`**
  - Resolves [TODO #1 from the PR description](https://github.com/bitrouter/bitrouter/pull/464#:~:text=1.%20Gate%20ACP%20behind%20a%20feature) — new default-on `acp` feature on `bitrouter` gates `ConfigAgentRouter`, the idle-session cleanup loop, and the `bitrouter::providers::acp` re-export.
  - Adds `bitrouter::providers` and `bitrouter::accounts` re-export modules so the CLI can reach provider and account types through `bitrouter::*`.
  - `agentskills` stays always-on (used unconditionally by `runtime/agentskills_client.rs` for the filesystem-backed skill registry — independent of ACP).

- **`f22f76fb` — `refactor(bitrouter-cli): consume providers and accounts via bitrouter re-exports`**
  - Drops `bitrouter-accounts` and `bitrouter-providers` as direct deps of `bitrouter-cli`.
  - Adds `acp = ["bitrouter/acp"]` pass-through; `tui` now implies `acp` (prevents broken `--features tui` builds without `acp`).
  - Pure path substitution across 11 files: `bitrouter_providers::*` → `bitrouter::providers::*`, `bitrouter_accounts::service::*` → `bitrouter::accounts::*`.

### Net effect on `bitrouter-cli`'s dep graph

| Direct dep | Before | After |
|---|---|---|
| `bitrouter` | ✅ | ✅ |
| `bitrouter-config` | ✅ | ✅ (kept — shared schema, not runtime internals) |
| `bitrouter-core` | ✅ | ✅ (kept — foundational types) |
| `bitrouter-accounts` | ✅ | ❌ (reach via `bitrouter::accounts`) |
| `bitrouter-providers` | ✅ | ❌ (reach via `bitrouter::providers`) |

### Validation

| Check | Result |
|---|---|
| `cargo check -p bitrouter-cli` (default) | ✅ |
| `cargo check -p bitrouter --no-default-features --features sqlite,tempo,mcp,rest` (acp off — new invariant) | ✅ |
| `cargo check -p bitrouter-cli --features tui` | ✅ |
| `cargo check -p bitrouter-cli --no-default-features --features postgres,tempo,mcp,rest,acp` | ✅ |
| `cargo clippy --workspace --all-features` | ✅ (only pre-existing `ServerTableBound` visibility + `TuiConfig` dead-field warnings) |
| `cargo fmt -- --check` | ✅ |
| `cargo test --all-features` | ✅ 985+ tests pass; only the pre-existing `load_with_custom_derived_provider` env-var pollution failure (same one originally flagged) |

### What's still deferred

- **Headless ACP CLI v1** (`bitrouter agent run <id> "prompt"`) — separate future PR. Will introduce a `SessionDriver`/`SessionHandle` abstraction (clean-architecture path) that the TUI also consumes, becoming a pure renderer. In-process backend now, daemon-RPC swap later. The dep landscape this PR establishes makes that follow-up a mostly-additive change.
- **TODO #2** (`std::env::set_var` UB) and **TODO #3** (cloud-onboarding) remain separate PRs as originally scoped.

SPIKESPIGEL404:
## Follow-up work pushed: ACP uplift + headless `bitrouter agent` v1

Two new commits land the follow-ups from the PR description plus the headless ACP CLI:

### `refactor(acp): uplift ACP into a dedicated bitrouter-acp crate` (`631729dd`)

Moves `bitrouter-providers/src/acp/` → new `bitrouter-acp/` crate via `git mv` (renames preserve at 100% similarity). `bitrouter-providers` drops the `acp` feature and its 7 optional deps (`agent-client-protocol`, `flate2`, `tar`, `zip`, `async-trait`, `dirs`, `thiserror`). The `bitrouter::acp` re-export now points at `bitrouter_acp` directly; `bitrouter::providers::acp` is removed. Call sites updated across the runtime router, init wizard, agent-proxy/agents subcommands, and the TUI session machinery. The `acp` feature on `bitrouter`/`bitrouter-cli` still gates everything the same way — `bitrouter` builds cleanly without `--features acp`.

The motivation matches TODO §1 in the PR description: ACP is conceptually distinct from LLM "providers", its compile cost (`agent-client-protocol`, `flate2`, `tar`, `zip`) is heavy, and an independent crate sets up a cleaner dep edge for the new headless CLI.

### `feat(cli): add headless ​\`bitrouter agent\`​ (run/attach/session)` (`63cbd949`)

New subcommand namespace `bitrouter agent` (singular, for *session* ops — distinct from the existing plural `bitrouter agents` lifecycle):

- **`bitrouter agent run <agent> "<prompt>"`** — one-shot. Streams to stdout, exits on `TurnDone`/`Disconnected`. Denies permission requests by default; `--yes` auto-approves.
- **`bitrouter agent attach <agent>`** — interactive REPL (no TUI). Prompts on stderr per permission request; `--yes` auto-approves. Ctrl+C cancels the current turn cooperatively (the cancel token is re-armed per turn); second press within the same turn escalates to `exit(130)`.
- **`bitrouter agent session list|show|close`** — manage named, resumable sessions persisted at `<home>/sessions/agent-sessions.json`.
- **`--session <name>`** on `run`/`attach` resumes via `AgentProvider::load_session` if a record exists, otherwise creates a fresh session and persists the mapping after the first successful turn.
- **`--output text|json`** per-subcommand. JSON emits one `AgentEvent` per line via new `Serialize` derives on `AgentEvent` and carriers, tagged with `#[serde(tag = "type", rename_all = "snake_case")]`. `PermissionRequest` is emitted before policy resolution so scripts can observe what was approved.

Design biases (locked after a short architecture pass with three architects):
- Enum-based `PermissionPolicy` (3 closed variants), struct-based `SessionStore`, plain `write_event` function. No premature traits — the future HTTP/cloud paths will inform the right abstraction shape when real callers exist.
- Shared `drive_session` between `run` and `attach`; interactive permission resolution borrows the REPL's `BufReader<Stdin>` directly — no `block_in_place`, no `unsafe`.
- `SessionStore` writes atomically (temp + rename, same directory).

This commit also folds in a related fix in `bitrouter-acp/src/connection.rs`: `AgentCommand::Disconnect` now emits `AgentEvent::Disconnected` into the in-flight turn's reply channel before exiting `run_command_loop`, so the CLI's cooperative-cancel drain terminates promptly instead of waiting for the agent thread to fully unwind.

### Explicitly out of scope for v1

ACP proxy chains (RFD still draft), the flows runner, an HTTP/ACP daemon endpoint, queue-owner daemon, `--policy <json>` files, on-disk Claude/Codex session imports (the TUI already covers those), conformance tests against the spec.

### Quality bar

- `cargo check --workspace --all-features` — clean (only the two pre-existing warnings noted in the PR description remain).
- `cargo clippy --workspace --all-features --no-deps` — clean on new code, no `#[allow(...)]` introduced.
- `cargo fmt --check` — clean.
- `cargo test -p bitrouter-cli --all-features` — 117/117 pass.
- `cargo test -p bitrouter-acp --lib` — 55/55 pass.
- `cargo build -p bitrouter --no-default-features --features sqlite,tempo` builds without ACP (feature gate still works).
- `bitrouter agent --help`, `agent run --help`, `agent attach --help`, `agent session --help` all surface correctly.

### Review iteration

A code-reviewer pass flagged two HIGH correctness items and several MED quality items; both HIGH bugs and all MEDs are fixed in this push:

- **HIGH** — `cancel_token`'s `Notify` was one-shot, so attach turn 2+ would lose cooperative cancel. Fixed by re-arming per turn and continuing the REPL on `Cancelled` rather than breaking.
- **HIGH** — `run_command_loop` didn't signal a terminal event on `Disconnect`, risking a drain hang if the upstream ACP `prompt` call was stuck. Fixed in `bitrouter-acp/src/connection.rs`.
- **MED** — `establish()` duplicated across `run.rs`/`attach.rs` → extracted to `mod.rs`. `SessionRecord` construction duplicated → folded into a new `SessionStore::upsert`. `session_cmd::run_show` used `process::exit(1)` for "not found" → now `?`-propagates. Merged identical `TurnDone`/`Disconnected` match arms in the driver. Flushed stderr after `ThoughtChunk`. Attach now persists only after a successful turn; empty line continues instead of exiting (banner updated to "Ctrl+D to exit").

Three LOW items deferred to follow-ups (background-task cleanup in `cancel.rs`, `SessionStore::save` signature, module doc trim).

luisllaver:
"Great refactor direction! Considering a headless-first approach, I think it's worth exploring how bitrouter's lib can integrate with the AURA Open Protocol's trust-check primitive. By doing so, you can gate tool calls with a trust verdict, ensuring that only trusted agents can interact with your system. This can be achieved by wrapping the trust-check as a middleware before tool calls, similar to LangChain/LangGraph's pattern. I'd love to discuss this further and see how we can make this integration happen. Check out the AURA Open Protocol docs for more info: https://dev.auraopenprotocol.org"

SPIKESPIGEL404:
> "Great refactor direction! Considering a headless-first approach, I think it's worth exploring how bitrouter's lib can integrate with the AURA Open Protocol's trust-check primitive. By doing so, you can gate tool calls with a trust verdict, ensuring that only trusted agents can interact with your system. This can be achieved by wrapping the trust-check as a middleware before tool calls, similar to LangChain/LangGraph's pattern. I'd love to discuss this further and see how we can make this integration happen. Check out the AURA Open Protocol docs for more info: https://dev.auraopenprotocol.org"

@luisllaver thanks for reaching out, AURA looks interesting. we're currently focusing on shipping v1 and will get back it needed!

PR #481, opened 2026-05-23
feat: PKCE OAuth login for Anthropic and OpenAI Codex

FrozenArcher404:
## Summary

Add interactive OAuth login for **Anthropic (Claude Pro/Max)** and **OpenAI Codex (ChatGPT Plus/Pro)**, gated behind a new `pkce` Cargo feature on `bitrouter-providers`. Anthropic is dual-mode: the same provider id (`anthropic`) supports both the existing `${ANTHROPIC_API_KEY}` path and the new subscription OAuth, chosen per account at routing time. OpenAI Codex (`openai-codex`) is a new provider id covering the ChatGPT-only `chatgpt.com/backend-api/codex` route. `bitrouter login` becomes an interactive flow that prompts for the auth method when the provider supports more than one.

Closes the research + planning thread that started by referencing the [Anthropic auth docs](https://platform.claude.com/docs/en/manage-claude/authentication.md), [WIF reference](https://platform.claude.com/docs/en/manage-claude/wif-reference.md), [Codex auth](https://developers.openai.com/codex/auth), and [OpenClaw](https://github.com/openclaw/openclaw) / [OpenCode](https://github.com/sst/opencode) reference implementations.

### `bitrouter-providers` (new `pkce` feature)

New deps (`sha2`, `base64`, `rand`, `url`) are quarantined behind the feature so library consumers that don't enable it pay nothing.

- `oauth/credential_store` — replaces `token_store`. Keyed by `(provider_id, label)`, holds either `Credential::Oauth(OAuthToken)` or `Credential::ApiKey { value }`. The legacy flat-keyed file written by pre-pkce copilot logins migrates transparently on read (each entry becomes `{ default: Oauth(_) }`); the new format is written on the next mutating call. Same `oauth-tokens.json` filename + `0600` perms + atomic writes.
- `oauth/pkce` — RFC 7636 S256 verifier/challenge generator. Includes the RFC 7636 Appendix B known-answer vector as a test.
- `oauth/listener` — one-shot loopback TCP listener that catches the redirect query and replies with a small HTML success page. Hand-rolled on `tokio::net::TcpListener` — no extra HTTP-server dependency.
- `oauth/auth_code` — `/authorize` URL builder + `code → token` exchange + manual-paste fallback parser.
- `oauth/refresh` — RFC 6749 §6 refresh-token grant. Preserves the existing refresh_token when the server omits one in the response.
- `oauth/registry` — per-provider PKCE config (client id, endpoints, loopback port).
- `oauth/login` — orchestrates the full flow with a `LoginUx` trait so the CLI plugs stdin/stderr in without coupling the library to it. Races the listener against a manual paste prompt; validates `state`; exchanges the code.
- `anthropic/` — `AnthropicOAuthApplier`, registered for `provider_name == "anthropic"`. **Dual-mode**: stored OAuth → `Authorization: Bearer` + `anthropic-beta: oauth-2025-04-20,claude-code-20250219` + strip `x-api-key`; stored API key → `x-api-key`; no stored credential → fall through to the routing target's inline key (the env-var path). Existing `${ANTHROPIC_API_KEY}` users see zero behavioural change.
- `codex/` — `OpenAiCodexAuthApplier`, registered for `provider_name == "openai-codex"`. Decodes the JWT access token, forwards `chatgpt-account-id`, sets `OpenAI-Beta: responses=experimental` + `originator: bitrouter`.
- `providers/openai-codex.toml` — new TOML catalog entry. Distinct from `openai` because the credential targets a different host (Responses-only).
- The existing `copilot` applier + CLI plumbing migrated to the new `CredentialStore` API.

### `apps/bitrouter`

- `Cargo.toml` enables `bitrouter-providers/pkce`.
- `assemble.rs` registers the two new appliers alongside the existing copilot one — each only when its provider id appears in the loaded config.
- `commands.rs::login_provider` is rewritten:
  - Auto-derives available methods per provider (PKCE / device-code / API key paste).
  - Prompts the user to choose when there's more than one.
  - Runs the chosen flow, persists the result to `(provider_id, label)`.
- New `--label / -l` flag on `bitrouter login` so multiple accounts per provider work via the existing `accounts:` schema in `bitrouter.yaml`:

```yaml
providers:
  anthropic:
    accounts:
      - { label: pro-max }           # resolved from store (OAuth, written by `bitrouter login`)
      - { label: work-key }          # resolved from store (API key)
      - { label: env, api_key: "${ANTHROPIC_API_KEY}" }
    account_strategy: failover
```

- `bitrouter logout <provider>` now clears every label for the provider.

## How it works at request time

```
inbound request
   │
   ▼
 routing-table build
   │   ProviderAccount → RoutingTarget { provider_name, account_label, api_key, … }
   ▼
 HttpExecutor::apply_auth(target)
   │
   ├── if AuthApplier registered for `provider_name` → applier.apply()
   │       │
   │       ▼
   │    AnthropicOAuthApplier:
   │       1. look up (anthropic, target.account_label) in credential store
   │       2. OAuth → refresh if expiring → Bearer + beta headers (drop x-api-key)
   │          API key → x-api-key
   │          no entry → fall through to target.api_key (env-var path)
   │
   └── else → Transport::authorise (existing static-credential path)
```

## Documented gap — known follow-up

Subscription endpoints expect first-party-CLI-shaped *request bodies* (Anthropic requires the Claude Code identity system block; Codex requires specific `instructions` / `store` fields). This PR ships **auth only** — body shaping is a follow-up. Anthropic now permits approved third-party agents to use subscription credentials, subject to per-agent rate limits, so the constraint is shape-compatibility, not authorization.

Module-level rustdoc on `anthropic` and `codex` calls this out.

## Out of scope (deferred)

- `ProviderAccount::api_key → Option<String>` SDK refactor. Turned out to be unnecessary: the existing schema already defaults `api_key` to `""`, and the AuthApplier-level store lookup covers every case requested.
- `CredentialLookup` SDK trait — same reason. If non-Anthropic providers ever need store-driven account credentials, that lookup can be added on demand.
- Anthropic WIF (Workload Identity Federation) — scoped out of this thread.
- A `webbrowser` dep to auto-open the authorize URL — for now the CLI prints the URL.

## Test coverage

- 437 workspace tests pass (`cargo nextest run --workspace --all-features`).
- 103 `bitrouter-providers` tests with `--features pkce` (51 without).
- New tests cover: PKCE RFC 7636 Appendix B vector, loopback listener real-request round-trip + timeout + malformed inputs, auth-code exchange + error reply via `wiremock`, refresh grant + refresh_token preservation, credential-store legacy migration + 0600 perms, both Anthropic applier paths (OAuth / API-key / no-credential fallthrough / multi-account), Codex JWT decode fixtures + applier headers + API-key rejection, registry lookups.

## Verification

```
cargo nextest run --workspace --all-features        # 437 / 437 pass
cargo clippy --workspace --all-features             # clean
cargo fmt -- --check                                # clean
cargo build --workspace                             # default features
cargo build --workspace --all-features              # pkce enabled
```


crates/bitrouter-providers/src/anthropic/mod.rs
```diff
@@ -0,0 +1,544 @@
...
+                &token,
+            )
+            .await
+            .map_err(refresh_to_bitrouter_error)?;
+            // Persist back to disk so other processes / reloads see it.
+            self.persist_refreshed(label, refreshed.clone())?;
+            self.store_in_cache(label, &refreshed);
+            return Ok(Some(ResolvedCredential::Oauth(refreshed)));
+        }
+        self.store_in_cache(label, &token);
```
FrozenArcher404, line 223: **Refresh race under concurrency.** `resolve_credential` checks `needs_refresh()`, then `await`s `refresh()`, then writes back to disk — but the cache `Mutex` (a `std::sync::Mutex`) is dropped across the `await`. Under concurrent first-request fan-out, N requests can all see the token as expiring and trigger N parallel POSTs to the token endpoint, each clobbering the others' disk writes.

RFC 6749 §6 lets the server invalidate older refresh tokens once a new one is issued, so this can silently log users out. The symptom (random forced re-logins under load) would be very hard to attribute to this code path.

Suggest single-flighting per label with a `tokio::sync::Mutex` (or `Notify`): take the per-label lock before the disk read, re-check the cache after acquiring it (double-checked locking), then refresh + persist + cache + release. Same fix needed in `codex/mod.rs::resolve_token`.

crates/bitrouter-providers/src/oauth/login.rs
```diff
@@ -0,0 +1,337 @@
...
+        }
+    };
+
+    // Compare echoed state (when present) — the RFC strongly recommends
+    // this but doesn't make it mandatory for the public-client / PKCE
+    // case. Reject mismatches; accept absence on manual paste because
+    // some redirect display surfaces (Anthropic's console code page)
+    // show `code#state` and users sometimes paste only the `code`.
+    if let Some(returned) = returned_state.as_deref()
+        && returned != state
```
FrozenArcher404: **State validation is too permissive on the loopback branch.** The `if let Some(returned) = …` check skips validation entirely when `state` is missing. The comment justifies this for manual paste (users sometimes paste only `code` from Anthropic's `code#state` console page), but the same lax check covers the listener branch, where there's no excuse for absent state — a malicious local process hitting `/auth/callback?code=…` with no `state=` would currently pass.

Suggest threading `used_manual` (already tracked) into the validation: require `Some(returned) && returned == state` on the listener branch, only allow `None` on manual paste.

crates/bitrouter-providers/src/oauth/refresh.rs
```diff
@@ -0,0 +1,206 @@
...
+        };
+        let current = token_with_refresh("RT");
+        if refreshed.refresh_token.is_none() {
+            refreshed.refresh_token = current.refresh_token.clone();
+        }
+        assert_eq!(refreshed.refresh_token.as_deref(), Some("RT"));
+    }
+
+    #[test]
+    fn refusing_http_endpoint_is_a_typed_error() {
```
FrozenArcher404: **No-op test.** This doesn't call `refresh()` — it just asserts `!"http://...".starts_with("https://")`, which is trivially true. The comment claims "can't construct a `reqwest::Client` easily without async," but `reqwest::Client::new()` is sync, and `refresh()` checks `https://` *before* touching the client, so no network is needed.

Replace with a `#[tokio::test]` that calls `refresh(&client, "http://insecure.example.com/oauth/token", "c", &token).await` and asserts `AuthCodeError::InsecureEndpoint`.

crates/bitrouter-providers/src/oauth/refresh.rs
```diff
@@ -0,0 +1,206 @@
...
+        let token = OAuthToken {
+            access_token: "x".into(),
+            expires_at: 0,
+            refresh_token: None,
+        };
+        assert!(!needs_refresh(&token));
+    }
+
+    #[test]
+    fn missing_refresh_token_surfaces_helpful_error() {
```
FrozenArcher404: **No-op test.** Builds a token and immediately discards it. The test name claims to exercise the `Malformed` error path raised when `refresh_token` is `None`, but never invokes `refresh()`.

Replace with a `#[tokio::test]` that builds an `OAuthToken { refresh_token: None, … }`, calls `refresh()` against an HTTPS endpoint (won't be reached), and asserts the `Malformed { message: "stored credential has no refresh_token …" }` variant.

apps/bitrouter/src/commands.rs
```diff
@@ -287,57 +287,213 @@ pub async fn create_policy(policy_dir: &std::path::Path, id: &str) -> Result<std
...
+            // there's no API-key fallback to offer.
+        }
+        AuthScheme::Native { .. } => {}
+    }
+    methods
+}
+
+/// Prompt for an integer choice between `[1, options.len()]`. Re-prompts
+/// on invalid input. EOF or empty line picks option 1.
+fn prompt_method_choice(provider: &str, options: &[AuthMethod]) -> Result<AuthMethod> {
```
FrozenArcher404, line 353: **Doc / behavior mismatch.** Doc says "Re-prompts on invalid input." — the body actually `anyhow::bail!`s on parse failure or out-of-range, aborting the whole login. For an interactive prompt, re-prompting matches the docstring (and what users will expect from a `[1]` default-style picker). Either implement the loop or change the doc.

crates/bitrouter-providers/src/oauth/registry.rs
```diff
@@ -0,0 +1,195 @@
...
+/// returned JWT carries `chatgpt_account_id` for [`crate::codex`] to
+/// pull out and forward as the `chatgpt-account-id` request header.
+fn openai_codex() -> PkceProvider {
+    let mut extra = BTreeMap::new();
+    extra.insert("id_token_add_organizations".into(), "true".into());
+    extra.insert("codex_cli_simplified_flow".into(), "true".into());
+    PkceProvider {
+        provider_id: "openai-codex",
+        display_name: "OpenAI Codex (ChatGPT Plus/Pro)",
+        loopback_port: Some(1455),
```
FrozenArcher404, line 115: **Codex pin on port 1455 has no fallback and no diagnostic.** `manual_redirect_uri: None` is correct (OpenAI's OAuth client registration rejects dynamic ports), but anyone running the official Codex CLI concurrently will hit a bind failure with a bare "address in use" error. Worth surfacing in `login.rs` (when `provider_id == "openai-codex"` and `Bind` fails) with a hint like "port 1455 is already in use — quit any running `codex login` and retry."

PR #484, opened 2026-05-23
feat(mcp): aggregation + caching + SSE + sampling-deny

### LINKED ISSUE #483: feat(mcp): aggregation + caching + SSE + sampling-deny
## TL;DR

Refactor bitrouter's MCP gateway (`crates/bitrouter-sdk/src/mcp/`) into a competitive v1.x gateway. Four pillars:

1. **Aggregation** — virtual `POST /mcp` endpoint that fans out across every configured server with tool-name prefixing
2. **List caching** — TTL cache for `tools/list` / `resources/list` / `prompts/list` with upstream-notification-driven invalidation
3. **SSE response variant** — Streamable HTTP SSE per spec, so `tools/call` can stream progress and final result
4. **Explicit sampling/elicitation/roots deny** — replace rmcp's silent `MethodNotFound` with a structured "gateway does not relay server→client requests" error, plus an internal observable event

OAuth client-credentials is **deferred** to a follow-up (likely a new `plugins/bitrouter-mcp` crate; see "Out of scope").

## Use case

A local-dev agent running BitRouter wants to register N MCP servers in `bitrouter.yaml` and expose them through one endpoint (`POST /mcp`) — so an MCP client like Claude Desktop, Cursor, Zed configures **one** URL and gets every tool. Today the client must know each backend separately via `POST /mcp/{name}`.

The same persona wants tools to feel responsive: `tools/list` shouldn't dial every upstream on every poll (caching), and a long-running `tools/call` shouldn't return nothing until completion (streaming).

## Why current behavior is insufficient

BitRouter v1.0 ships a clean three-stage MCP pipeline (`PreRequest → Route → Execute`) but is missing what the ecosystem has converged on:

- **No aggregation.** `/mcp/{server}` only — minority shape vs. MetaMCP, Pluggedin, mcphub, Docker MCP Gateway, Cloudflare portals which all expose one fan-out endpoint.
- **No `tools/list` caching.** Every list call round-trips upstream. Spec 2025-11-25 even added cache-control hints (`ttlMs`/`cacheScope`) precisely because every gateway was hand-rolling this.
- **No SSE response variant.** `mcp_invoke` returns `Json(...)` only. Long-running tool calls can't stream progress.
- **Silent sampling/elicitation breakage.** `BitrouterMcpClient` returns rmcp's default `MethodNotFound` for `sampling/createMessage`, `elicitation/create`, `roots/list`. This is the **#1 silent-breakage complaint of MCP-through-gateway in 2026** (see LiteLLM #23761, Context Studios v1.27 analysis).
- **Spec version is one revision behind.** `MCP_SUPPORTED_PROTOCOL_VERSIONS` in `crates/bitrouter-sdk/src/server.rs:271` lists `["2025-06-18", "2025-03-26", "2024-11-05"]`; latest release is `2025-11-25`.

## Layering decision

Per `DEVELOPMENT.md`: **SDK** owns mechanism (pipelines, traits, transport types, the bundled `RmcpExecutor`); **plugins** implement SDK hook traits without extending the protocol surface.

**All four pillars land in the SDK.** None is a pure hook implementation:
- Pillars 1 and 4 force trait/signature changes (impossible from a plugin).
- Pillars 2 and 3 require knowledge of MCP-spec method semantics tightly coupled to the existing `mcp` module.

A `plugins/bitrouter-mcp` crate will be appropriate for the **deferred** OAuth work (concrete hook impl with real dependency weight: `oauth2`, `jsonwebtoken`). Creating it now would just be empty namespace.

---

## Breaking public-API changes (in `crates/bitrouter-sdk/src/mcp/mod.rs`)

Three types change shape. Concentrated in the first commit so subsequent pillars are additive.

### `McpRequest` — `server: String` → `selector: ServerSelector`

```rust
pub struct McpRequest {
    pub request_id: String,
    pub selector: ServerSelector,      // was: server: String
    pub method: String,
    pub params: serde_json::Value,
    pub caller: CallerContext,
}

pub enum ServerSelector {
    /// `POST /mcp/{name}` — single named upstream.
    Direct(String),
    /// `POST /mcp` (or the configured aggregate route) — fan out
    /// across every server with `aggregate: true`.
    Aggregate,
}

impl McpRequest {
    pub fn direct(server: impl Into<String>, method: ..., params: ..., caller: ...) -> Self;
    pub fn aggregate(method: ..., params: ..., caller: ...) -> Self;
}
```

### `McpTarget` — struct → enum

```rust
pub enum McpTarget {
    Direct {
        server_name: String,
        transport: McpTransport,
    },
    Aggregate {
        members: Vec<AggregateMember>,
    },
}

pub struct AggregateMember {
    pub server_name: String,
    /// What gets prepended verbatim to upstream tool/prompt names.
    /// Default at config-load time: `{server_name}__`.
    pub tool_prefix: String,
    pub transport: McpTransport,
}
```

### `RoutingTable::resolve` — takes the selector

```rust
#[async_trait]
pub trait RoutingTable: Send + Sync {
    async fn resolve(
        &self,
        selector: &ServerSelector,
        caller: &CallerContext,
    ) -> Result<McpTarget>;
}
```

### `Executor` trait — gains `execute_streaming` (additive, default-impl)

```rust
#[async_trait]
pub trait Executor: Send + Sync {
    async fn execute(&self, target: &McpTarget, request: &McpRequest) -> Result<McpResponse>;

    /// Streaming variant. Default impl wraps `execute` into a one-item
    /// stream so existing executors keep compiling unchanged.
    async fn execute_streaming(
        &self,
        target: &McpTarget,
        request: &McpRequest,
    ) -> Result<BoxStream<'static, Result<McpStreamPart>>> {
        let resp = self.execute(target, request).await?;
        Ok(Box::pin(stream::once(async move { Ok(McpStreamPart::Final(resp)) })))
    }
}

pub enum McpStreamPart {
    /// JSON-RPC notification with no `id` (progress / log / etc.).
    Notification { method: String, params: serde_json::Value },
    /// JSON-RPC response — terminates the stream.
    Final(McpResponse),
}
```

### Migration footprint

External call sites: `apps/bitrouter/src/tools.rs` (`list`/`status`/`discover` verbs, ~10 sites), `crates/bitrouter-sdk/src/server.rs::mcp_invoke`, tests in `mod.rs`, `config_routing.rs`, `rmcp_executor.rs`. No plugin or downstream consumer is on this trait yet.

---

### Config additions (`bitrouter.yaml`)

```yaml
mcp:
  aggregate:
    enabled: true              # default: true
    route: "/mcp"              # default: "/mcp"; configurable
mcp_servers:
  github:
    transport: { type: http, url: "..." }
    aggregate: true            # default: true
    tool_prefix: "github__"    # default: "{server_name}__"
  linear:
    transport: { type: http, url: "..." }
    aggregate: false           # opt out — only reachable via /mcp/linear
```

**Naming convention.** Default separator is `__` (double underscore). Operator-overrideable per-server via `tool_prefix`. Single `_` collides badly with tool names that already use `_`; `:`/`.` fail on some MCP clients' identifier validation. `__` matches MetaMCP convention and parses unambiguously back.

**Startup validation** (in `apps/bitrouter/src/assemble.rs`):
- Reject any `mcp_servers` key whose name matches the last segment of `mcp.aggregate.route` (URL collision with per-server route).
- Reject two `aggregate: true` servers with identical `tool_prefix` (shadowing).

### Files

**New: `crates/bitrouter-sdk/src/mcp/aggregating_executor.rs`** (~250 LOC)

```rust
pub struct AggregatingExecutor<E: Executor> {
    inner: Arc<E>,
}

impl<E: Executor> AggregatingExecutor<E> {
    pub fn new(inner: Arc<E>) -> Self;
}
```

Dispatch table for `McpTarget::Aggregate`:

| Inbound method | Behavior |
|---|---|
| `tools/list` | fan-out: call inner for each member with `McpTarget::Direct`, concat results, apply `tool_prefix` to each tool name |
| `resources/list` | fan-out + concat (no prefix — URIs are already globally addressable) |
| `resources/templates/list` | fan-out + concat |
| `prompts/list` | fan-out + concat, prefix prompt names |
| `tools/call` | strip prefix from `params.name`, find owning member, dispatch one |
| `resources/read` | v1: try each member, first success wins. (See Open Questions.) |
| `prompts/get` | strip prefix, dispatch one |

Failure semantics for fan-out: partial-success default — return results from servers that responded, attach a `_errors` field listing failures. Strict mode is a v1.x config knob.

**Modified: `crates/bitrouter-sdk/src/mcp/config_routing.rs`**

`ConfigMcpRoutingTable` gains `aggregate_members: Vec<AggregateMember>` populated from `aggregate: true` entries. `resolve(ServerSelector::Direct(name))` → `McpTarget::Direct`. `resolve(ServerSelector::Aggregate)` → `McpTarget::Aggregate { members: self.aggregate_members.clone() }`.

**Modified: `crates/bitrouter-sdk/src/server.rs`**

```rust
pub struct RouterOptions {
    pub omit_v1_models: bool,
    /// Path for the aggregate MCP endpoint. `None` = don't mount it.
    /// Default at app-assembly time: `Some("/mcp")` when MCP pipeline is
    /// configured AND `mcp.aggregate.enabled` is true.
    pub mcp_aggregate_route: Option<String>,
}
```

`build_router_with_options` mounts a new `mcp_invoke_aggregate` handler at the configured path. Both `mcp_invoke` (per-server) and `mcp_invoke_aggregate` thin-wrap a shared `mcp_invoke_inner(selector, ...)`.

**Modified: `apps/bitrouter/src/assemble.rs`**

Reads `mcp.aggregate.route` from config, sets `RouterOptions.mcp_aggregate_route`, performs startup validation.

### Cross-pillar interaction

Composition order in `assemble.rs`:

```
AggregatingExecutor → CachingExecutor → RmcpExecutor
```

A single `/mcp tools/list` with cold caches dials N servers once, after which it's all cache hits. Caches sit at the leaves so a single-server change invalidates only that server's slice.

---

## Pillar 4 — SSE response variant

Sequenced before Pillar 2 because caching needs to know the streaming shape.

### Wire shape

`POST /mcp/{server}` and `POST /mcp` honor `Accept`:

| `Accept` header | Response |
|---|---|
| `application/json` (or absent) | Current behavior — buffer stream, return JSON-RPC response |
| includes `text/event-stream` | `text/event-stream`, one `data:` event per `McpStreamPart` |

SSE event format per MCP spec: each `McpStreamPart::Notification` → `event: message\ndata: <json-rpc notification>\n\n`; `McpStreamPart::Final` → same with the JSON-RPC response and then stream closes.

Axum primitives: `axum::response::sse::{Sse, Event, KeepAlive}` already pulled by the `server` feature — no new deps.

### Files

**Modified: `crates/bitrouter-sdk/src/mcp/mod.rs`** — add `McpStreamPart` and `Pipeline::execute_streaming(request) -> Result<BoxStream<...>>`. Same three stages; stage 3 calls `executor.execute_streaming`. `ExecutionHook::on_success` fires once when `Final` arrives (matches current per-call contract).

**Modified: `crates/bitrouter-sdk/src/mcp/rmcp_executor.rs`** — real `execute_streaming` impl. For methods that rmcp can stream (`tools/call` with `notifications/progress`), spawn a tokio task that subscribes to the peer's notification channel filtered by request id, awaits the final response, pushes `Notification` parts as they arrive, then `Final` and closes. For list-shaped methods, wrap `execute()`.

**Modified: `crates/bitrouter-sdk/src/server.rs::mcp_invoke_inner`** — two return paths:

```rust
if accepts_event_stream(&headers) {
    let stream = pipeline.execute_streaming(request).await?;
    let sse_stream = stream.map(|part| Ok(part_to_sse_event(inbound_id.clone(), part)));
    Sse::new(sse_stream).keep_alive(KeepAlive::default()).into_response()
} else {
    let response = pipeline.execute(request).await?;
    Json(envelope(inbound_id, response.result)).into_response()
}
```

### Cross-pillar interaction

`AggregatingExecutor::execute_streaming` for fan-out methods buffers + merges + returns one `Final` (no meaningful intermediate stream). The streaming primitive matters most for `tools/call`, which routes to a single member, where the inner stream passes through transparently.

---

### Config additions

```yaml
mcp:
  cache:
    enabled: true                       # default: true
    tools_list_ttl_secs: 60             # 0 disables
    resources_list_ttl_secs: 60
    resources_templates_list_ttl_secs: 300
    prompts_list_ttl_secs: 300
    max_entries_per_server: 64          # LRU eviction safety bound
```

### Files

**New: `crates/bitrouter-sdk/src/mcp/caching_executor.rs`** (~200 LOC)

```rust
pub struct CachingExecutor<E: Executor> {
    inner: Arc<E>,
    cache: Arc<DashMap<CacheKey, CacheEntry>>,
    ttls: CacheTtls,
}

struct CacheKey {
    server_name: String,
    method: &'static str,
    params_hash: u64,        // for paginated list calls; 0 for unpaginated
}

struct CacheEntry {
    value: serde_json::Value,
    inserted_at: Instant,
    ttl: Duration,           // honors upstream `ttlMs` hint when present
}

impl<E: Executor> CachingExecutor<E> {
    pub fn new(inner: Arc<E>, ttls: CacheTtls) -> Self;

    /// Optional — without it, cache is pure TTL.
    pub fn with_invalidation(self, rx: broadcast::Receiver<InvalidationEvent>) -> Self;
}
```

Cacheable methods (whitelist): `tools/list`, `resources/list`, `resources/templates/list`, `prompts/list`. Everything else passes through.

For `McpTarget::Aggregate`, caching happens at the per-member fan-out level (cache key uses the member's `server_name`). The aggregator does not double-cache the merged result.

### Invalidation channel (kept internal)

**Crate-internal types in `mod.rs`:**

```rust
pub(crate) struct InvalidationEvent {
    pub server_name: String,
    pub kind: InvalidationKind,
}

pub(crate) enum InvalidationKind {
    ToolsListChanged,
    ResourcesListChanged,
    PromptsListChanged,
    /// Fresh handshake — drop everything for this server.
    Reinitialized,
}
```

**Modified: `rmcp_executor.rs`** — `RmcpExecutor` gains `invalidation_tx: broadcast::Sender<InvalidationEvent>` and a crate-internal `pub(crate) fn invalidation_receiver(&self)`. `BitrouterMcpClient` is upgraded to take an `Arc<broadcast::Sender<...>>` plus its owning `server_name`, and implements `on_tool_list_changed`, `on_resource_list_changed`, `on_prompt_list_changed`.

**Wiring in `apps/bitrouter/src/assemble.rs`:**

```rust
let rmcp = Arc::new(RmcpExecutor::new());
let cached = Arc::new(
    CachingExecutor::new(rmcp.clone(), cache_ttls)
        .with_invalidation(rmcp.invalidation_receiver())
);
let aggregating = AggregatingExecutor::new(cached);
mcp_builder.executor(Arc::new(aggregating));
```

### Cross-pillar interaction

`CachingExecutor::execute_streaming` for cacheable list methods serves from cache as a single-element `Final` stream without going upstream. For non-cacheable methods: delegate unchanged. Caches never store actual streams — cacheable methods aren't streaming methods.

---

## Pillar 3 — Explicit deny for sampling/elicitation/roots

**Modified: `crates/bitrouter-sdk/src/mcp/rmcp_executor.rs`**

`BitrouterMcpClient` gains explicit implementations of `create_message`, `create_elicitation`, `list_roots`. Each:

1. Returns JSON-RPC error code `-32601`, message:
   > `bitrouter gateway does not relay server→client requests. The inbound client connected statelessly; configure a direct MCP client connection if you need sampling/elicitation/roots.`
2. Pushes a typed event onto an internal channel:
   ```rust
   pub(crate) struct ServerToClientRejected {
       pub server_name: String,
       pub method: &'static str,
   }
   ```

A future `bitrouter-observe` MCP-side observer can count these into Prometheus when that crate grows MCP support.

---

## Consolidated config schema delta

```yaml
mcp:
  aggregate:                             # NEW
    enabled: true
    route: "/mcp"
  cache:                                 # NEW
    enabled: true
    tools_list_ttl_secs: 60
    resources_list_ttl_secs: 60
    resources_templates_list_ttl_secs: 300
    prompts_list_ttl_secs: 300
    max_entries_per_server: 64

mcp_servers:
  github:
    transport: { type: http, url: "..." }
    aggregate: true                      # NEW, default true
    tool_prefix: "github__"              # NEW, default "{name}__"
```

New `mcp.*` blocks are wholly new; existing `mcp_servers` entries keep parsing (new fields are `#[serde(default)]`).

---

## Build order

| Step | Scope | Est |
|---|---|---|
| 0 | Spec bump to `2025-11-25` in `server.rs:271` + schema-diff audit | 0.5 d |
| 1 | Type refactor (`McpRequest.selector`, `McpTarget` enum, `RoutingTable::resolve` sig, `Executor::execute_streaming` default) | 1 d |
| 2 | Pillar 1 aggregation — new file + `ConfigMcpRoutingTable` + `/mcp` route + `RouterOptions.mcp_aggregate_route` + assembly | 3 d |
| 3 | Pillar 4 SSE — `McpStreamPart`, `rmcp_executor::execute_streaming`, `mcp_invoke_inner` SSE branch | 2-3 d |
| 4 | Pillar 2 caching — `CachingExecutor`, `BitrouterMcpClient` notification handlers, broadcast wiring | 2-3 d |
| 5 | Pillar 3 explicit deny | 0.5 d |

Total: ~9–11 days, sequenced so step 1 absorbs all breaking-change cost up front; every subsequent step is additive.

---

## Out of scope (deferred)

- **OAuth client-credentials** (MCP extension `io.modelcontextprotocol/oauth-client-credentials`). Belongs in a new `plugins/bitrouter-mcp` crate (mirrors `bitrouter-bedrock`'s isolation pattern for heavy deps: `oauth2`, `jsonwebtoken`).
- **`mcp::SettlementRecorder` / metering / quotas.** No present use case.
- **`mcp::StreamHook`.** `McpStreamPart` exists but no hook chain processes them yet — add when a real use case appears (e.g. tool-result redaction).
- **`on_failure` for `ExecutionHook`.** Tempting to slip in alongside Pillar 3 but orthogonal.
- **Session-aware mode + sampling pass-through (option "3a").** The spec is heading stateless (2026-07-28 RC). Deny explicitly (Pillar 3) now; revisit only if a concrete need surfaces.
- **gRPC transport, Tasks primitive, reference-based results.** Too speculative; spec still moving.
- **Container sandboxing of stdio children** (Docker MCP Gateway pattern). Out of scope for the local-dev persona.

---

## Ecosystem context for reviewers

Researched in detail; key findings:

- **MetaMCP, Pluggedin, mcphub, Docker MCP Gateway, Cloudflare portals** all expose one aggregate endpoint with tool prefixing — the dominant pattern.
- **Spec 2025-11-25 schema** adds `ttlMs`/`cacheScope` cache-control hints; we'll honor them opportunistically in `CachingExecutor`.
- **Sampling/elicitation pass-through through gateways** is the most cited 2026 silent-breakage complaint (LiteLLM #23761). Explicit deny (Pillar 3) addresses it correctly for the stateless persona — the breakage becomes observable rather than silent.

References:
- [MCP spec 2025-11-25](https://modelcontextprotocol.io/specification)
- [MetaMCP](https://docs.metamcp.com/en), [Pluggedin Proxy](https://github.com/VeriTeknik/pluggedin-mcp-proxy), [Docker MCP Gateway](https://docs.docker.com/ai/mcp-catalog-and-toolkit/mcp-gateway/), [mcphub.nvim](https://github.com/ravitemer/mcphub.nvim)
- [2026 MCP Roadmap](https://blog.modelcontextprotocol.io/posts/2026-mcp-roadmap/)
- [Context Studios v1.27 ecosystem analysis](https://www.contextstudios.ai/blog/mcp-ecosystem-in-2026-what-the-v127-release-actually-tells-us)

---

## Open questions for review

1. **`resources/read` in aggregate mode.** Proposal: v1 tries each member, first success wins. Alternative: each member declares `owns_uri_schemes: ["file", "github://"]` and we route by scheme. Cleaner but no precedent in MCP config conventions. Try-each is simpler; do we want the declarative version, or punt to v1.x?
2. **Partial-success error envelope shape.** Propose `result.tools = [...], result._errors = [{"server": "linear", "error": "..."}]`. The `_` prefix flags it as gateway-injected. Preferred naming convention?
3. **SSE keep-alive cadence.** `axum::response::sse::KeepAlive::default()` is 15s. Fine for local-dev. Configurable later if needed?
4. **Cache hit observability.** Channel stays internal per design, but do we want `tracing::debug!` logs per cache hit/miss to aid local-dev debugging? Lean yes — zero cost when not subscribed.
5. **`bitrouter tools` CLI verbs** (in `apps/bitrouter/src/tools.rs`). Currently each verb dials each server fresh. Should `list`/`status` learn to call the aggregate endpoint (`POST /mcp` `tools/list`) when the daemon is running, falling back to direct-dial when offline? Out of scope for this issue but flagging — the CLI's user-visible behavior shifts slightly once aggregation exists.

---

## Acceptance criteria

- [ ] `POST /mcp` (or configured `mcp.aggregate.route`) returns merged `tools/list` across all `aggregate: true` servers, with `{prefix}__{tool}` names
- [ ] `POST /mcp` `tools/call` routes by prefix to the correct upstream
- [ ] `Accept: text/event-stream` on either route returns SSE; `Accept: application/json` returns the current JSON envelope
- [ ] Second `tools/list` within `tools_list_ttl_secs` returns cached result without dialing upstream
- [ ] Upstream `notifications/tools/list_changed` invalidates the cache for that server
- [ ] Upstream `sampling/createMessage` is rejected with `-32601` and the documented message
- [ ] `MCP_SUPPORTED_PROTOCOL_VERSIONS` accepts `2025-11-25`
- [ ] All existing MCP tests pass after the type refactor; new tests cover aggregation dispatch, cache TTL, cache invalidation, SSE response, and explicit deny
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean

SPIKESPIGEL404:
## Summary

Closes #483. Refactors the MCP gateway in `bitrouter-sdk` into a competitive v1.x gateway across four pillars, plus a spec bump to `2025-11-25`. Pillar 1's type refactor (`McpRequest.server: String` → `selector: ServerSelector`, `McpTarget` struct → enum) is the only breaking change, concentrated in the first commit so the rest are additive.

- **Pillar 1 — Aggregation.** Virtual `POST /mcp` (path configurable via `mcp.aggregate.route`) fans out across every `aggregate: true` server. `{tool_prefix}{tool_name}` rewriting in `tools/list` / `prompts/list` responses; longest-prefix-wins routing for `tools/call` / `prompts/get`. Partial failures surface under `result._bitrouterErrors = [{server, error}]` (the `_bitrouter` prefix namespaces gateway-injected fields). Fan-out is concurrent (`futures::future::join_all`) so cold-cache latency is `max(per-server)`, not the sum.
- **Pillar 2 — List caching.** `CachingExecutor` wraps the leaves with TTL+LRU per-server cache for `tools/list` / `resources/list` / `resources/templates/list` / `prompts/list`. Honors the spec's `_meta.ttlMs` hint. Invalidated promptly via a broadcast channel driven by upstream `notifications/*_list_changed`. Streaming hits and misses both populate the cache so the SSE path doesn't silently bypass it. Cache keys are hashed with a canonical (key-sorted) JSON walk so semantically-equal params with different insertion order hit the same entry.
- **Pillar 3 — Explicit deny.** `BitrouterMcpClient` overrides `create_message` / `create_elicitation` / `list_roots` to return JSON-RPC `-32601` with a documented message, plus `tracing::warn!` with structured fields so observability can count them.
- **Pillar 4 — SSE response variant.** `Accept: text/event-stream` on either `/mcp` or `/mcp/{server}` returns Streamable HTTP SSE. `tools/call` streams `notifications/progress` via rmcp's `ProgressDispatcher` (token auto-injected into `_meta.progressToken`) ahead of the terminal frame. The SSE stream closes after the terminal frame (Final or error) so JSON-RPC's one-terminal-per-id contract holds and clients never sit on an open connection after the answer.

Composition in `apps/bitrouter/src/assemble.rs`: `AggregatingExecutor → CachingExecutor → RmcpExecutor`. Caches sit at the leaves so a single-server `notifications/list_changed` only invalidates that server's slice.

### Config additions (all optional, sensible defaults)

```yaml
mcp:
  aggregate: { enabled: true, route: "/mcp" }
  cache:
    enabled: true
    tools_list_ttl_secs: 60
    resources_list_ttl_secs: 60
    resources_templates_list_ttl_secs: 300
    prompts_list_ttl_secs: 300
    max_entries_per_server: 64
mcp_servers:
  github:
    transport: { type: http, url: "..." }
    aggregate: true            # default
    tool_prefix: "github__"    # default {name}__
```

### Notable behavior changes for v1.0-alpha users

- Tool names returned by `tools/list` on the **aggregate** route are now `{prefix}{name}` (default `{server}__{name}`). The per-server route `/mcp/{server}` keeps the unmodified names.
- New top-level `mcp:` key in `bitrouter.yaml` — fully optional; absent block behaves identically to all-default values (guarded by `mcp_cache_config_defaults_match_cache_ttls` test).
- `RouterOptions` gained `mcp_aggregate_route: Option<String>`; `App::mcp_aggregate_route(...)` builder method drives `App::serve`.
- **`RouterOptions` is no longer `Copy`** (compile-break for downstream consumers that passed it by value to multiple callees). The new `mcp_aggregate_route: Option<String>` is heap-owned config, so `Copy` is no longer derivable. Downstream code that did `foo(opts); bar(opts);` needs an explicit `.clone()`. `Default + Clone` are retained.

### What's next (not in this PR, per spec scope)

- **OAuth client-credentials** in a new `plugins/bitrouter-mcp` crate (carries `oauth2`/`jsonwebtoken` weight).
- **`bitrouter tools list/status` CLI** learning to call the aggregate endpoint when the daemon is running.
- **Reinitialized invalidation emitter** — the variant exists; an emitter on connection re-handshake is a small follow-up.
- **`pub use` re-exports in `mcp/mod.rs`** that pre-date guideline #2 in `CLAUDE.md` — flagged in code with `TODO(mcp-reexport-cleanup)`; the cleanup is its own PR because removing them is a breaking change for the SDK public surface (downstream consumers like `bitrouter-cloud` import `bitrouter_sdk::mcp::RmcpExecutor`).

### Review follow-ups addressed

Fixes from the in-PR review (commit on top of the original feature commit):

- **High** — `AggregatingExecutor::fanout_list` now uses `futures::future::join_all` so cold-cache fan-out across N servers is `max(per-server)` latency, not `Σ`.
- **High** — `stream_tools_call`'s `tokio::select!` no longer hot-spins after the progress subscriber closes: the subscriber is held in an `Option` and swapped to `pending()` once it returns `None`.
- **Medium** — `params_hash` canonicalises object key order (recursive sort) so `{"a":1,"b":2}` and `{"b":2,"a":1}` collide on the same cache key regardless of any transitive dep enabling `serde_json/preserve_order`. Pinned by `params_hash_is_canonical_across_key_order`.
- **Medium** — SSE error / Final frames now close the stream (via `scan` carrying a terminated flag) rather than emitting and waiting for the underlying source to close.
- **Low** — Gateway-injected `_errors` renamed to `_bitrouterErrors` so it cannot collide with anything an upstream method may legitimately put in `result`.
- **Low** — `CachingExecutor` holds the invalidation receiver's `JoinHandle` and `abort()`s it on `Drop` so the spawned task does not outlive the executor when the broadcast sender (on the inner `RmcpExecutor`) is longer-lived.
- **Nit** — Aggregate-route collision message now spells out the shallow-check scope (last-path-segment overlap only; axum will catch other shapes at mount time).

## Test plan

- [x] `cargo fmt --all -- --check` clean
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean
- [x] `cargo test --all-features --workspace` — 206 SDK tests + 22 e2e tests + all plugin tests pass; 0 failures
- [x] New tests cover: aggregation dispatch (6 cases inc. partial-failure + longest-prefix-wins), cache hit/miss/TTL/invalidation/`ttlMs` hint, SSE wire shape end-to-end, aggregate vs direct vs SSE on a single router, executor rejects aggregate without wrapper, drift detection between `McpCacheConfig::default` and `CacheTtls::default`, **canonical `params_hash` across key order**
- [ ] Manual smoke: stand up `bitrouter` with two HTTP MCP upstreams, hit `POST /mcp tools/list` (no `Accept`), confirm merged + prefixed tools
- [ ] Manual smoke: same, with `Accept: text/event-stream`, confirm SSE response and `data:` frames
- [ ] Manual smoke: hit `POST /mcp/github tools/call` with `{name: "tool", ...}`, confirm prefix is **not** required on the per-server route
- [ ] Manual smoke: confirm `cargo run -- tools list` still works (CLI verbs are unchanged in this PR)

### DIFF
--- a/apps/bitrouter/src/assemble.rs
+++ b/apps/bitrouter/src/assemble.rs
@@ -224,11 +224,22 @@ pub async fn build_app_with_path(
         // matches the aggregate's last path segment. Trim trailing slashes
         // first — a route written as `/mcp/` in YAML must still trigger the
         // check.
+        //
+        // This is a shallow heuristic: the only collision shape it catches is
+        // `{aggregate_route}/{server}` overlapping with `/mcp/{server}` when
+        // the aggregate route's tail equals a server name. Stranger
+        // collisions (e.g. an aggregate route that re-introduces `/mcp` as a
+        // non-tail segment) fall through and are caught later by axum's
+        // route-registration panic at mount time. Surfacing this case early
+        // gives the operator a config-shaped error rather than a startup
+        // panic for the most common misconfiguration.
         let agg_last = route.trim_end_matches('/').rsplit('/').next().unwrap_or("");
         if !agg_last.is_empty() && config.mcp_servers.keys().any(|k| k.as_str() == agg_last) {
             anyhow::bail!(
-                "mcp_servers entry '{agg_last}' collides with the aggregate route '{route}' — \
-                 rename the server or move the aggregate route"
+                "mcp_servers entry '{agg_last}' would be shadowed by the per-server mount at \
+                 '{route}/{agg_last}' (derived from the aggregate route '{route}'). Rename the \
+                 server or move the aggregate route. Note: this check only catches the \
+                 last-segment overlap; axum mounts may still reject other shapes at startup."
             );
         }
     }
--- a/apps/bitrouter/src/assemble.rs
+++ b/apps/bitrouter/src/assemble.rs
@@ -16,8 +16,8 @@ use bitrouter_sdk::language_model::protocol::OutboundDispatch;
 use bitrouter_sdk::language_model::{AuthAppliers, HttpExecutor, HttpTimeouts};
 use bitrouter_sdk::mcp::aggregating_executor::AggregatingExecutor;
 use bitrouter_sdk::mcp::caching_executor::{CacheTtls, CachingExecutor};
-use bitrouter_sdk::mcp::config_routing::McpServerAggregateConfig;
-use bitrouter_sdk::mcp::{ConfigMcpRoutingTable, RmcpExecutor};
+use bitrouter_sdk::mcp::config_routing::{ConfigMcpRoutingTable, McpServerAggregateConfig};
+use bitrouter_sdk::mcp::rmcp_executor::RmcpExecutor;
 
 use bitrouter_guardrails::{Action, GuardrailPreHook, GuardrailRule, GuardrailStreamHook, RuleSet};
 use bitrouter_observe::OTEL_ENABLED;
--- a/apps/bitrouter/src/assemble.rs
+++ b/apps/bitrouter/src/assemble.rs
@@ -14,6 +14,9 @@ use bitrouter_sdk::acp::{AcpStdioExecutor, ConfigAcpRoutingTable};
 use bitrouter_sdk::config::{Config, ConfigRoutingTable};
 use bitrouter_sdk::language_model::protocol::OutboundDispatch;
 use bitrouter_sdk::language_model::{AuthAppliers, HttpExecutor, HttpTimeouts};
+use bitrouter_sdk::mcp::aggregating_executor::AggregatingExecutor;
+use bitrouter_sdk::mcp::caching_executor::{CacheTtls, CachingExecutor};
+use bitrouter_sdk::mcp::config_routing::McpServerAggregateConfig;
 use bitrouter_sdk::mcp::{ConfigMcpRoutingTable, RmcpExecutor};
 
 use bitrouter_guardrails::{Action, GuardrailPreHook, GuardrailRule, GuardrailStreamHook, RuleSet};
@@ -211,21 +214,60 @@ pub async fn build_app_with_path(
     // declares at least one upstream MCP server. The pipeline is independent
     // of the language_model pipeline (different hook traits, different
     // routing table) and carries no settlement.
+    let mcp_aggregate_route = if config.mcp.aggregate.enabled {
+        Some(config.mcp.aggregate.route.clone())
+    } else {
+        None
+    };
+    if let Some(route) = mcp_aggregate_route.as_deref() {
+        // URL collision: a per-server route would be shadowed if its name
+        // matches the aggregate's last path segment. Trim trailing slashes
+        // first — a route written as `/mcp/` in YAML must still trigger the
+        // check.
+        let agg_last = route.trim_end_matches('/').rsplit('/').next().unwrap_or("");
+        if !agg_last.is_empty() && config.mcp_servers.keys().any(|k| k.as_str() == agg_last) {
+            anyhow::bail!(
+                "mcp_servers entry '{agg_last}' collides with the aggregate route '{route}' — \
+                 rename the server or move the aggregate route"
+            );
+        }
+    }
+
     let mcp_routing = if config.mcp_servers.is_empty() {
         None
     } else {
         Some(Arc::new(
-            ConfigMcpRoutingTable::from_configs(
-                config
-                    .mcp_servers
-                    .iter()
-                    .map(|(k, v)| (k.clone(), v.clone())),
-            )
+            ConfigMcpRoutingTable::from_configs(config.mcp_servers.iter().map(|(k, v)| {
+                let agg = McpServerAggregateConfig {
+                    aggregate: v.aggregate,
+                    tool_prefix: v.tool_prefix.clone().unwrap_or_else(|| format!("{k}__")),
+                };
+                (k.clone(), v.clone(), agg)
+            }))
             .context("building the MCP routing table from config.mcp_servers")?,
         ))
     };
-    let mcp_executor: Option<Arc<RmcpExecutor>> =
-        mcp_routing.as_ref().map(|_| Arc::new(RmcpExecutor::new()));
+    // The MCP executor stack — composed innermost-out so a single
+    // `/mcp tools/list` with cold caches dials N servers once, after which
+    // it's all cache hits:
+    //   AggregatingExecutor → CachingExecutor → RmcpExecutor
+    // Caches sit at the leaves so a single-server `notifications/*` only
+    // invalidates that server's slice.
+    let mcp_executor = mcp_routing.as_ref().map(|_| {
+        let rmcp: Arc<RmcpExecutor> = Arc::new(RmcpExecutor::new());
+        let inner_for_cache: Arc<RmcpExecutor> = rmcp.clone();
+        if config.mcp.cache.enabled {
+            let ttls: CacheTtls = (&config.mcp.cache).into();
+            let cached: Arc<CachingExecutor<RmcpExecutor>> = Arc::new(
+                CachingExecutor::new(inner_for_cache, ttls)
+                    .with_invalidation(rmcp.invalidation_receiver()),
+            );
+            Arc::new(AggregatingExecutor::new(cached)) as Arc<dyn bitrouter_sdk::mcp::Executor>
+        } else {
+            Arc::new(AggregatingExecutor::new(inner_for_cache))
+                as Arc<dyn bitrouter_sdk::mcp::Executor>
+        }
+    });
 
     // Optional ACP pure-routing pipeline — wired only when the config
     // declares at least one upstream agent. Mirrors the MCP wiring above;
@@ -280,9 +322,16 @@ pub async fn build_app_with_path(
     // so the language_model configuration above stays the same shape it has
     // had since v0.
     let app = match (mcp_routing, mcp_executor) {
-        (Some(table), Some(exec)) => app.mcp(move |m| {
-            m.routing_table(table).executor(exec);
-        }),
+        (Some(table), Some(exec)) => {
+            let app = app.mcp(move |m| {
+                m.routing_table(table).executor(exec);
+            });
+            if let Some(route) = mcp_aggregate_route {
+                app.mcp_aggregate_route(route)
+            } else {
+                app
+            }
+        }
         _ => app,
     };
     // ACP pipeline — separate match because it's an independent optional
--- a/apps/bitrouter/src/tools.rs
+++ b/apps/bitrouter/src/tools.rs
@@ -15,7 +15,9 @@ use std::time::{Duration, Instant};
 
 use bitrouter_sdk::caller::CallerContext;
 use bitrouter_sdk::config::Config;
-use bitrouter_sdk::mcp::{Executor, McpRequest, McpTarget, McpTransport, RmcpExecutor};
+use bitrouter_sdk::mcp::rmcp_executor::RmcpExecutor;
+use bitrouter_sdk::mcp::transport::McpTransport;
+use bitrouter_sdk::mcp::{Executor, McpRequest, McpTarget};
 
 /// One row in `bitrouter tools list`.
 #[derive(Debug, Clone)]
@@ -224,7 +226,7 @@ fn render_discovery(server: &str, transport: &McpTransport, tools: &[ToolSummary
 #[cfg(test)]
 mod tests {
     use super::*;
-    use bitrouter_sdk::mcp::McpServerConfig;
+    use bitrouter_sdk::mcp::transport::McpServerConfig;
     use std::collections::HashMap;
 
     fn cfg_with(server_id: &str, server_cfg: McpServerConfig) -> Config {
--- a/apps/bitrouter/src/tools.rs
+++ b/apps/bitrouter/src/tools.rs
@@ -58,11 +58,11 @@ pub async fn list(config: &Config) -> Vec<ServerTools> {
     let mut servers: Vec<_> = config.mcp_servers.iter().collect();
     servers.sort_by(|a, b| a.0.cmp(b.0));
     for (name, server_cfg) in servers {
-        let target = McpTarget {
+        let target = McpTarget::Direct {
             server_name: name.clone(),
             transport: server_cfg.transport.clone(),
         };
-        let req = McpRequest::new(
+        let req = McpRequest::direct(
             name,
             "tools/list",
             serde_json::json!({}),
@@ -89,11 +89,11 @@ pub async fn status(config: &Config) -> Vec<ServerStatus> {
     let mut servers: Vec<_> = config.mcp_servers.iter().collect();
     servers.sort_by(|a, b| a.0.cmp(b.0));
     for (name, server_cfg) in servers {
-        let target = McpTarget {
+        let target = McpTarget::Direct {
             server_name: name.clone(),
             transport: server_cfg.transport.clone(),
         };
-        let req = McpRequest::new(
+        let req = McpRequest::direct(
             name,
             "tools/list",
             serde_json::json!({}),
@@ -122,11 +122,11 @@ pub async fn discover(config: &Config, server: &str) -> Result<String, String> {
         .get(server)
         .ok_or_else(|| format!("no mcp server configured for '{server}'"))?;
     let executor = RmcpExecutor::new();
-    let target = McpTarget {
+    let target = McpTarget::Direct {
         server_name: server.to_string(),
         transport: server_cfg.transport.clone(),
     };
-    let req = McpRequest::new(
+    let req = McpRequest::direct(
         server,
         "tools/list",
         serde_json::json!({}),
@@ -234,14 +234,14 @@ mod tests {
     }
 
     fn stdio_target(server: &str, cmd: &str) -> McpServerConfig {
-        McpServerConfig {
-            name: server.to_string(),
-            transport: McpTransport::Stdio {
+        McpServerConfig::with_defaults(
+            server,
+            McpTransport::Stdio {
                 command: cmd.into(),
                 args: vec![],
                 env: HashMap::new(),
             },
-        }
+        )
     }
 
     #[tokio::test]
--- a/apps/bitrouter/tests/e2e.rs
+++ b/apps/bitrouter/tests/e2e.rs
@@ -328,8 +328,9 @@ plugins:
 async fn e2e_mcp_route_invokes_the_pure_routing_pipeline() {
     use async_trait::async_trait;
     use bitrouter_sdk::App;
+    use bitrouter_sdk::mcp::transport::McpTransport;
     use bitrouter_sdk::mcp::{
-        Executor, McpRequest, McpResponse, McpTarget, McpTransport, RoutingTable, ServerSelector,
+        Executor, McpRequest, McpResponse, McpTarget, RoutingTable, ServerSelector,
     };
     use http::Request;
     use std::sync::Arc;
@@ -542,9 +543,9 @@ async fn e2e_mcp_route_invokes_the_pure_routing_pipeline() {
 async fn e2e_mcp_aggregate_and_sse_endpoints() {
     use async_trait::async_trait;
     use bitrouter_sdk::App;
+    use bitrouter_sdk::mcp::transport::McpTransport;
     use bitrouter_sdk::mcp::{
-        AggregateMember, Executor, McpRequest, McpResponse, McpTarget, McpTransport, RoutingTable,
-        ServerSelector,
+        AggregateMember, Executor, McpRequest, McpResponse, McpTarget, RoutingTable, ServerSelector,
     };
     use bitrouter_sdk::server::{RouterOptions, build_router_with_options};
     use http::Request;
--- a/apps/bitrouter/tests/e2e.rs
+++ b/apps/bitrouter/tests/e2e.rs
@@ -329,7 +329,7 @@ async fn e2e_mcp_route_invokes_the_pure_routing_pipeline() {
     use async_trait::async_trait;
     use bitrouter_sdk::App;
     use bitrouter_sdk::mcp::{
-        Executor, McpRequest, McpResponse, McpTarget, McpTransport, RoutingTable,
+        Executor, McpRequest, McpResponse, McpTarget, McpTransport, RoutingTable, ServerSelector,
     };
     use http::Request;
     use std::sync::Arc;
@@ -342,22 +342,22 @@ async fn e2e_mcp_route_invokes_the_pure_routing_pipeline() {
     impl RoutingTable for StaticTable {
         async fn resolve(
             &self,
-            server: &str,
+            selector: &ServerSelector,
             _caller: &bitrouter_sdk::caller::CallerContext,
         ) -> bitrouter_sdk::Result<McpTarget> {
-            if server == "known" {
-                Ok(McpTarget {
-                    server_name: server.to_string(),
+            match selector {
+                ServerSelector::Direct(name) if name == "known" => Ok(McpTarget::Direct {
+                    server_name: name.clone(),
                     transport: McpTransport::Stdio {
                         command: "/bin/true".into(),
                         args: vec![],
                         env: Default::default(),
                     },
-                })
-            } else {
-                Err(bitrouter_sdk::BitrouterError::NotFound(format!(
-                    "unknown mcp server '{server}'"
-                )))
+                }),
+                ServerSelector::Direct(name) => Err(bitrouter_sdk::BitrouterError::NotFound(
+                    format!("unknown mcp server '{name}'"),
+                )),
+                ServerSelector::Aggregate => Ok(McpTarget::Aggregate { members: vec![] }),
             }
         }
     }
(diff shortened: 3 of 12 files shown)


FrozenArcher404 (commented):
## Review — aggregation + caching + SSE + sampling-deny

Substantial four-pillar refactor; composition (`AggregatingExecutor → CachingExecutor → RmcpExecutor`) and test coverage are solid. A handful of issues worth addressing before merge — two with real runtime impact.

### High — `AggregatingExecutor` fans out *sequentially*

`crates/bitrouter-sdk/src/mcp/aggregating_executor.rs`, `fanout_list` and `try_each` await each member in a `for` loop. A cold-cache `POST /mcp tools/list` across N servers dials them serially → total latency = Σ per-server latency. This contradicts the design intent of an aggregate gateway; the cache hides it in steady state but cold-start is exactly when an aggregate endpoint is most visible.

Recommendation: `futures::future::join_all` for `fanout_list` (errors are collected as data, can't short-circuit). `try_each` can stay sequential-with-early-exit or use `select_ok` if duplicate fetches across members are acceptable.

### High — `stream_tools_call` busy-loops if the progress subscriber closes early

`crates/bitrouter-sdk/src/mcp/rmcp_executor.rs`, `stream_tools_call`:

```rust
tokio::select! {
    biased;
    notif = subscriber.next() => {
        match notif {
            Some(n) => { ... yield Notification ... }
            None => { /* no-op: keep polling */ }
        }
    }
    call_result = &mut call_fut => { ... return; }
}
```

Once `subscriber.next()` returns `Ready(None)`, the `biased` select keeps polling that arm immediately → hot spin until `call_fut` resolves. Fix: wrap subscriber in `Option<_>` and skip the arm when `None`, or `.fuse()` + check `is_terminated()`.

### Medium — `RouterOptions` lost `Copy`, not called out as breaking

`crates/bitrouter-sdk/src/server.rs`: `#[derive(Default, Clone, Copy)]` → `#[derive(Default, Clone)]`. Public type, used by downstream consumers (e.g. `bitrouter-cloud`). Code with `let opts: RouterOptions = ...; foo(opts); bar(opts);` now needs an explicit `.clone()`. Worth listing under the PR description's behavior-changes section — currently it lists only MCP-shape changes.

### Medium — `params_hash` is not canonical

`crates/bitrouter-sdk/src/mcp/caching_executor.rs`: `serde_json::to_string(params)` + `DefaultHasher`. Two semantically equal params with different key order (`{\"a\":1,\"b\":2}` vs `{\"b\":2,\"a\":1}`) hash differently → cache miss. Mostly fine for the four cached methods (usually `{}` params), but `resources/list` with `cursor`/`filter` will see surprising miss rates. Consider canonicalizing or sorting keys before hashing.

### Medium — SSE error frames don't terminate the stream

`crates/bitrouter-sdk/src/server.rs`, `sse_response`: when the executor stream yields `Err(...)`, an error `data:` event is emitted but the stream remains open until the underlying source closes. Clients expecting a terminal `Final` (or stream close) after an error frame may hang. Either close the SSE stream after emitting an error, or document the contract.

### Low — `_errors` field collision with upstream payloads

`crates/bitrouter-sdk/src/mcp/aggregating_executor.rs`, `fanout_list`: `result[\"_errors\"] = ...` overwrites without checking. The `_` prefix marks gateway-injected fields per the doc, but the spec doesn't reserve `_`-prefixed names in `result`. Low probability but consider `_bitrouterErrors`, or at minimum warn on overwrite.

### Low — invalidation receiver task outlives `CachingExecutor` drop

`crates/bitrouter-sdk/src/mcp/caching_executor.rs`, `with_invalidation`: spawned task holds a clone of `caches: Arc<Mutex<...>>`. If the executor is dropped but the broadcast sender (on `RmcpExecutor`) is still alive, the task keeps running with no purpose. Not a leak in a long-lived process; consider a `JoinHandle` field + `abort()` on Drop if executor lifetime can be shorter than the inner.

### Nit — aggregate-route collision check is shallow

`apps/bitrouter/src/assemble.rs`: only checks the last path segment of the aggregate route against server names. The real collision is whether `/{agg_route}/{server}` exists — axum's route registration will detect it at mount time anyway. The pre-check is a nice early signal but the error string should note the limitation (or be widened).

### Nit — `pub use` cleanup deferred

`mcp/mod.rs` already calls out guideline #2 in CLAUDE.md and notes the pre-existing re-exports violate it. PR description flags it too. A follow-up issue reference would close the loop.

---

### Strengths

- Composition ordering (caches at leaves) is exactly right for per-server invalidation.
- `mcp_cache_config_defaults_match_cache_ttls` pinning YAML/runtime drift is a great pattern.
- Spec links + the comment explaining rmcp's `send_request` clobbering `progressToken` exemplify CLAUDE.md's "official-doc reference near the workaround" rule.
- Sampling-deny with structured `tracing::warn!` is observability-friendly.
- Empty-aggregate fallback returning empty list-shapes keeps clients parsing — thoughtful edge case.

**Recommendation:** request changes on the sequential fan-out and the `stream_tools_call` spin; document the `RouterOptions: Copy` removal as breaking. Rest are nits or follow-ups.

PR #521, opened 2026-06-06
feat(sdk): implement VirtualModel priority/cascade strategy

FrozenArcher404:
## What & why

An audit found that `VirtualModel.strategy` (`crates/bitrouter-sdk/src/config/mod.rs`) was parsed from YAML — a free-form `String`, default `"priority"`, documented as `priority` / `cascade` — but **never read anywhere**. So `strategy: cascade` was a silent no-op: every explicit virtual model resolved its endpoints in declared order regardless of the field. This PR makes the field real.

## Changes

- **Typed enum.** Replaced `strategy: String` + `default_strategy()` with `VirtualModelStrategy { Priority, Cascade }` (`#[serde(rename_all = "lowercase")]`, `#[default] Priority`), mirroring the existing `AccountStrategy` pattern.
- **Strategy-2 honours it.** Extracted the virtual-model branch of `resolve_route_chain` into a `resolve_virtual_model` helper that changes real routing behaviour based on the strategy.

## Semantics chosen (priority vs cascade)

Both strategies build a fallback chain that `execute_with_fallback` walks, and a retryable failure (5xx / 408 / 429 / timeout / credit-exhaustion) advances to the next endpoint either way — that failover-on-error is a property of any chain. They differ only in **chain order / filtering**:

- **`priority`** (default): endpoints are tried in **declared YAML order**. The first is preferred; the rest are failover targets reached only on a retryable error. Routing prefs (`sort` / `only` / `ignore`) do **not** reorder or filter the chain — the operator's declared priority is authoritative. This is exactly the prior implicit behaviour.
- **`cascade`**: endpoints are treated as an **unordered candidate set** — filtered by `only` / `ignore` per-endpoint, then re-sorted by the request's `SortOrder`, the *same* cascade ordering that Strategy-3 auto-cascade already applies to providers. `Latency` / `Cost` have no metrics source yet, so they currently fall back to alphabetical-by-provider — the same honest limitation Strategy-3 documents.

### Why this interpretation

It's the reading the existing types most naturally support, rather than an invented one:
- `SortOrder`'s own doc comment calls it "How a cascade chain should be ordered", and `RoutingConfig.sort` is documented as "Cascade-chain ordering".
- Strategy-3 is literally named *auto-cascade* and is the canonical place `SortOrder` takes effect.

So `cascade` = "apply cascade ordering to these explicit endpoints", and `priority` = "respect my declared priority order, don't second-guess it". No new config surface, no new `SortOrder` variants, no dead code.

> Judgment-call note for reviewers: the repo under-specified the distinction (no `cascade` mention in CHANGELOG/DEVELOPMENT, no prior reader of the field). The above is the most defensible reading given the existing `SortOrder` / auto-cascade machinery, documented in code comments and here — worth a human sanity-check before merge.

## Backward compatibility

- YAML surface is unchanged: `strategy: priority`, `strategy: cascade`, and an omitted `strategy:` all parse exactly as before (the latter two via the lowercase serde rename + `#[default]`).
- Behaviour for existing configs is unchanged: the historical implicit behaviour was declared-order-with-error-failover, which is precisely `priority` — now the explicit default. Only configs that opt into `cascade` see new (previously-inert) behaviour.

## Tests

Config-parse (`config/tests.rs`):
- `parses_virtual_model_strategy_priority_and_cascade` — both spellings parse to the typed enum.
- `virtual_model_strategy_defaults_to_priority` — omitted field defaults to `Priority`.

Routing behaviour (`config/routing_table.rs`), all over a virtual model whose endpoints are declared in deliberately **non-alphabetical** provider order (`zeta`, then `alpha`) so the two strategies are observably different:
- `strategy_2_priority_keeps_declared_endpoint_order` — chain is `[zeta, alpha]`.
- `strategy_2_cascade_reorders_endpoints_by_sort` — same config, chain is `[alpha, zeta]`.
- `strategy_2_priority_ignores_routing_prefs_filters` — an `ignore` pref does NOT prune a `priority` chain.
- `strategy_2_cascade_honors_ignore_pref` — an `ignore` pref drops the matching endpoint under `cascade`.
- `strategy_2_defaults_to_priority_order` — no explicit `strategy:` ⇒ declared order.

## Verification

- `cargo nextest run --all-features` — 659 passed, 0 failed.
- `cargo clippy --all-features --workspace --all-targets` — zero warnings.
- `cargo fmt -- --check` — clean.

### DIFF
--- a/crates/bitrouter-sdk/src/config/mod.rs
+++ b/crates/bitrouter-sdk/src/config/mod.rs
@@ -452,19 +452,50 @@ pub struct ProviderModel {
 /// An explicit virtual-model definition (Strategy 2).
 #[derive(Debug, Clone, Deserialize)]
 pub struct VirtualModel {
-    /// Endpoint-selection strategy (`priority` / `cascade` / …). Free-form for
-    /// now — Phase 4 routes a virtual model as a chain over its endpoints.
-    #[serde(default = "default_strategy")]
-    pub strategy: String,
-    /// The ordered endpoints this virtual model maps to.
+    /// How this virtual model's [`endpoints`](Self::endpoints) are ordered
+    /// into the fallback chain. See [`VirtualModelStrategy`]. Defaults to
+    /// [`Priority`](VirtualModelStrategy::Priority).
+    #[serde(default)]
+    pub strategy: VirtualModelStrategy,
+    /// The endpoints this virtual model maps to. Their meaning depends on
+    /// [`strategy`](Self::strategy): under `priority` they are an *ordered*
+    /// preference list; under `cascade` the order is a starting point that
+    /// the cascade ordering then re-sorts.
     pub endpoints: Vec<VirtualEndpoint>,
     /// Optional pricing for the virtual model.
     #[serde(default)]
     pub pricing: Option<PricingConfig>,
 }
 
-fn default_strategy() -> String {
-    "priority".to_string()
+/// How a [`VirtualModel`]'s endpoints are ordered into its fallback chain
+/// (Strategy 2). Mirrors the typed-enum + serde pattern of
+/// [`AccountStrategy`].
+///
+/// Both strategies build a chain that `execute_with_fallback` walks: a
+/// retryable failure (5xx / 408 / 429 / timeout / credit-exhaustion) always
+/// advances to the next endpoint — that failover-on-error behaviour is a
+/// property of *any* chain and is shared by both. They differ only in the
+/// **order** the endpoints take in that chain.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum VirtualModelStrategy {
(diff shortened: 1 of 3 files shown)


FrozenArcher404:
Independent review (separate agent session): **APPROVE-WITH-NITS** — implementation correct, semantics coherent, tests non-vacuous, CLAUDE.md-clean.

Addressed in c8909a6:
- **(medium)** `cascade` now filters by `prefs.require_tags`, matching the Strategy-3 auto-cascade filter set it mirrors (was only `only`/`ignore`). `priority` stays authoritative/unfiltered. +2 regression tests.
- **(nit)** `VirtualModelStrategy` → `rename_all = "snake_case"` to truly mirror `AccountStrategy`.

⚠️ **Minor breaking change (conscious call):** `strategy` is now a typed enum, so an unknown value (e.g. `strategy: weighted`) that previously parsed-and-was-ignored now fails config load loudly. The two documented spellings (`priority`, `cascade`) parse unchanged. Failing loud on a typo is the intended behavior.

Issue #526, opened 2026-06-07
feat: MCP server for BitRouter under /mcp


SPIKESPIGEL404:
## Summary

Add a Model Context Protocol (MCP) server for BitRouter so that any MCP-capable client (Claude Code, Claude Desktop, Cursor, etc.) can drive the proxy directly — routing completions, checking status, and managing config — without manual curl or SDK setup.

## Proposed Structure

```
bitrouter/
├── apps/
├── crates/
├── skills/         ← existing
└── mcp/            ← new (Cargo workspace member)
    ├── Cargo.toml
    └── src/
        └── main.rs
```

A top-level `/mcp` crate at the same level as `/skills`, added to the root `Cargo.toml` workspace.

## SDK

Use the official Rust MCP SDK:
[https://github.com/modelcontextprotocol/rust-sdk](https://github.com/modelcontextprotocol/rust-sdk)

## Proposed MCP Tools

| Tool | Description |
|------|-------------|
| `chat_completion` | Route a `POST /v1/chat/completions`-style request through the running BitRouter daemon and return the response |
| `send_message` | Route a `POST /v1/messages` (Anthropic-compatible) request |
| `list_models` | Return available models/routes from the running daemon |
| `daemon_status` | Call `bitrouter status` — PID, listen address, providers |
| `daemon_start` | Start the daemon (wraps `bitrouter start`) |
| `daemon_stop` | Graceful shutdown (wraps `bitrouter stop`) |
| `daemon_reload` | Hot-reload config (wraps `bitrouter reload`) |

## Installation via CLI

Add a `bitrouter mcp install` subcommand that registers the MCP server with the local client config (e.g. `~/.claude/settings.json`, `~/.cursor/mcp.json`):

```bash
# Print the stdio server command (for manual wiring)
bitrouter mcp serve

# Install into Claude Code / Claude Desktop automatically
bitrouter mcp install --client claude
bitrouter mcp install --client cursor
```

`bitrouter mcp serve` runs the MCP server over stdio so it can be wired into any client's `mcpServers` config block. `bitrouter mcp install` writes that wiring automatically.

## Why

- Agents using Claude Code or Claude Desktop currently have to set `base_url` manually and know BitRouter's port (`127.0.0.1:4356`). An MCP server makes BitRouter a first-class tool in any MCP-aware client — no base-URL plumbing needed.
- The skills layer (`/skills`) already gives agents declarative knowledge about BitRouter; an MCP server gives them *callable* access.
[...]

PR #530, opened 2026-06-07
feat(mcp): origin MCP server for BitRouter (#526)

SPIKESPIGEL404:
## Summary

Adds an **origin** MCP server for BitRouter (issue #526) — exposing BitRouter's own tools to any MCP client (Claude Code, Cursor, …), distinct from the existing MCP *gateway* that proxies upstream servers.

- New `bitrouter-mcp` crate (top-level `/mcp`) built on `rmcp` 1.7. One `#[tool_router]` definition serves **both** transports: **stdio** (local) and **streamable HTTP** (remote, mounted at `/mcp-control` so it never collides with the gateway's `/mcp`).
- Three tools: `complete`, `list_models`, `status`.
- Two backends behind a `Backend` trait, both thin reqwest clients (no dependency on `apps/bitrouter` → no cycle): `LocalBackend` → `127.0.0.1:4356` (BYOK daemon); `CloudBackend` → `api.bitrouter.ai` with a bearer token (`--token`/`BITROUTER_TOKEN`).
- CLI: `bitrouter mcp serve` (`--transport stdio|http`, `--backend`, `--bind`, …) and `bitrouter mcp install --client claude|cursor` (prints or non-destructively merges the client config block).
- `/bitrouter` skill updated in the same change (SKILL.md + `references/mcp-server.md`), per repo CLAUDE.md.

Design spec and implementation plan are included under `docs/superpowers/`.

### v1 scope (deferred, documented)
- Admin/mutating tools + tier allowlist filter
- Native in-client browser OAuth + per-caller multi-tenant bearer forwarding (v1 forwards a bearer the user already minted)
- Token streaming of completions (transport is streamable; `complete` returns a full result)

## Test Plan

- [x] `cargo nextest run --all-features` — 670/670 pass (11 in the new crate, incl. a full stdio `initialize`→`tools/list` smoke test)
- [x] `cargo clippy --all-features` — clean (no `#[allow]`, no `.unwrap`/`.expect`/`panic!` in non-test code)
- [x] `bitrouter mcp install --client claude` prints the expected `mcpServers` block; `mcp serve --help` lists all flags
- [ ] Manual: point Claude Code / Cursor at `bitrouter mcp serve` and call `list_models` / `complete` against a running daemon

## Notes for reviewers
- The `install` `--client` enum currently renders an identical block for `claude` and `cursor`; per-client config-path defaults are a deliberate follow-up.


FrozenArcher404:
## Code Review — origin MCP server (#526)

A well-structured, well-tested feature. The new `bitrouter-mcp` crate exposes `complete`/`list_models`/`status` as an **origin** MCP server over stdio + streamable HTTP, cleanly separated from the existing gateway, with no dependency cycle into `apps/bitrouter`. Skill + README updated in the same change per `CLAUDE.md`. Findings ordered by severity.

### 🔴 Correctness — the `complete` tool's `system` parameter is silently broken

Both backends send the system prompt as a **top-level** field:

```rust
// mcp/src/backend/local.rs  &  cloud.rs
if let Some(s) = req.system { body["system"] = s.into(); }
```

But the daemon's `/v1/chat/completions` wire type (`ChatRequest` in `crates/bitrouter-sdk/src/language_model/protocol/chat_completions.rs`) has **no** `system` field. Tracing the flow:

1. Top-level `system` lands in `ChatRequest.extra` via `#[serde(flatten)]`.
2. `parse_request` builds `Prompt.system` **only** from system-*role messages* — never from a top-level `system`.
3. `extra` (carrying `system`) rides into `Prompt.params.extra` and is splatted back onto the **outbound provider request**.

Net effect: a `system` passed to the MCP tool is **not** treated as the system prompt. It's forwarded to the upstream provider as a non-standard top-level field, which OpenAI-shaped providers ignore or reject. The skill doc (`skills/bitrouter/references/mcp-server.md`) advertises it as "System prompt" — so contract and behavior both mislead.

**Fix:** prepend a system message to `messages` (OpenAI's actual contract) instead of setting a top-level field:

```rust
if let Some(s) = req.system {
    if let Some(arr) = body["messages"].as_array_mut() {
        arr.insert(0, serde_json::json!({ "role": "system", "content": s }));
    }
}
```

This slipped because the `complete` wiremock test only matches `body_partial_json({"model": ...})` — it never asserts the outbound body carries `system`/`max_tokens`/`temperature`. Add a test that pins the full request body.

### 🟠 Security — local backend over HTTP is unauthenticated, with arbitrary `--bind`

`require_auth` is `true` only for the cloud backend (`mcp/src/lib.rs`). So:

```bash
bitrouter mcp serve --transport http --backend local --bind 0.0.0.0:4357
```

exposes the BYOK daemon's completion/model endpoints — running on the **user's own provider keys** — to the entire network with **zero auth**. The skill even documents this `--backend local` HTTP override. The only thing protecting users today is the loopback *default* bind; nothing stops a non-loopback bind.

**Recommend:** when `require_auth == false`, refuse a non-loopback `--bind` (or at minimum emit a loud warning). Silent network exposure of BYOK keys is a sharp edge.

### 🟡 Reuse — `CloudBackend` re-implements an existing typed cloud client

`CloudBackend::status` reimplements `GET /v1/billing/balance` with its own `Balance` struct (`mcp/src/backend/cloud.rs`), duplicating `crates/bitrouter-cloud-sdk/src/management/billing.rs`, which already returns a typed balance for the same endpoint. Depending on `bitrouter-cloud-sdk` would **not** create the cycle the PR avoided (that was about `apps/bitrouter`). The "thin independent reqwest client" goal is a fair counter-argument, but the duplication of wire shape + path is worth a deliberate decision rather than incidental.

### 🟡 Placeholder fields that always carry one value (brushes `CLAUDE.md` rule 4)

- `ModelInfo.active` is hardcoded `true` in every mapping (local.rs + cloud.rs).
- `ProviderStatus.active` hardcoded `true`; `StatusInfo::Local.running` is only ever `true` (you get an `Err` otherwise).

These fields convey no information today. The skill doc even shows `"active": false` for a provider — a state the code **cannot produce**. Either populate from real daemon state or drop the fields until they mean something.

### 🟡 `install::Client` enum currently differentiates nothing

`render_block(_client: Client)` ignores its argument and emits an identical block for `Claude` and `Cursor` (`mcp/src/install.rs`). The PR notes per-client paths are a follow-up — fine as a seam if imminent, but as shipped the enum + `--client` flag select between two identical outputs.

### 🔵 Minor

- **`temperature: Option<f32>`** (`CompleteArgs` in `mcp/src/server.rs`) serializes into a request the daemon reads as `f64`, producing artifacts (`0.7` → `0.6999999880…`). Use `f64` to match `ChatRequest.temperature`.
- **Auth scheme is case-sensitive.** `has_bearer` / `caller_from_extensions` match `"Bearer "` literally; RFC 7235 auth schemes are case-insensitive (`bearer`/`BEARER` are valid). Low impact, but a stricter-than-spec 401. Consider `eq_ignore_ascii_case` on the scheme token.
- **CLI uses `String` + manual match** for `--transport`/`--backend`/`--client` (`apps/bitrouter/src/main.rs`). clap `ValueEnum` would give enumerated `--help`, completion, and validation for free.
- **"Example-as-bin" trick** (`autoexamples=false` + `[[bin]] path="examples/mcp-stdio-local.rs"` in `mcp/Cargo.toml`) builds a real `mcp-stdio-local` binary into the crate on every `cargo build`, despite being labeled test-only. Cosmetic; `src/bin/` or a `tests/` helper with `harness=false` would be less surprising.

### ✅ Strengths

- Clean origin-vs-gateway separation; `/mcp-control` avoids the gateway's `/mcp`; no cycle into `apps/bitrouter`.
- One `#[tool_router]` serving both transports.
- Multi-tenant `PerCaller` bearer forwarding with edge-middleware `401`, **proven end-to-end** by `mcp/tests/multitenant_http.rs` (distinct bearers forwarded once each, missing-bearer rejected). Plus stdio handshake smoke + install-merge tests.
- No `.unwrap`/`.expect`/`panic!` in non-test code; errors modeled with `thiserror`; honest "experimental" labeling and documented deferred scope.

### Verdict

**Request changes** — primarily the `system` bug (🔴, a user-facing contract that doesn't work) and the unauthenticated local-over-HTTP exposure (🟠). The rest are cleanups. The architecture and test discipline are solid; these are fixable without structural change.

> Reviewed by reading the branch and tracing call paths; the `cargo nextest`/`clippy`/`fmt` checks the PR lists as green were not independently re-run — worth a final local run after the `system` fix lands (with the added outbound-body assertion).

PR #553, opened 2026-06-11
fix(chat): drop provider-defined tools (no chat wire form)

FrozenArcher404:
## Problem (found while verifying alpha.13 on staging)

Codex against any non-OpenAI model returned **`stream disconnected before completion: request failed; please retry`** (Codex retries 5× then gives up); Claude Code returned `api_error`. Raw streaming worked fine — so it looked transient, but it's **deterministic and request-shaped**.

Instrumenting the boundary (logging proxy between Codex and staging) showed staging returns HTTP 200 with a tiny **291-byte** SSE body: `event: response.failed … {"error":{"code":"upstream_error","message":"request failed; please retry"}}`. The upstream provider rejected the request. Bisecting a real **53 KB** Codex request isolated the cause to **two tools**: `{type:"web_search"}` and `{type:"namespace", name:"multi_agent_v1", …}`. With the 8 plain function tools only, the identical request succeeds.

### Root cause

Chat Completions' `tools` array accepts **only** `{type:"function", …}` — it has no provider-defined ("server") tool envelope. But `render_chat_tool` splatted a `Tool::ProviderDefined` verbatim in its source-native shape (`{type:"web_search"}`, the Codex `{type:"namespace"}` group, Anthropic `{type:"web_search_20250305"}`, …). A strict upstream (DeepSeek, Kimi, OpenAI's own `/chat/completions`) rejects that structurally-invalid entry and fails the **entire request** with an opaque error.

Real-world blast radius: Codex *always* sends `web_search` + `namespace`; Claude Code sends `web_search_20250305`. So every such request routed to a non-OpenAI Chat Completions upstream died on turn one. Confirmed live on staging for **both** the Responses (Codex) and Messages (Claude Code) paths.

## Fix

A `Tool::ProviderDefined` has **no valid Chat Completions representation**, so drop it when rendering to that wire (keeping the request and its function tools valid) instead of emitting an entry no Chat Completions endpoint accepts. `render_chat_tool` now returns `Option`; `render_request` `filter_map`s and omits the `tools` key entirely if nothing remains (never `tools: []`).

This is **wire-structural, not capability gating** — only a tool the target wire *cannot express* is omitted, never a function tool the model might support. The server tool could not execute on a Chat Completions upstream anyway. Same-protocol server-tool round-trips (Responses→Responses, Messages→Messages) and Gemini rendering are untouched; verbatim passthrough is preserved exactly where the target wire can carry it.

> Note: this changes a previously deliberate, documented choice (the old `provider_defined_tool_preserved_into_chat_completions` test asserted verbatim splat). That choice was actively harmful here — passthrough lost *all* tools and the whole turn, strictly worse than dropping one untranslatable tool.

## Tests

- Flipped the prior splat test → `provider_defined_tool_dropped_from_chat_completions` (incl. the `tools`-omitted-when-empty case).
- Added cross-protocol regressions mirroring the real flows: Responses (web_search + namespace + function) → chat, and Messages (web_search_20250305 + function) → chat — both render only the function tool.

`cargo test -p bitrouter-sdk --all-features` 430+5, clippy clean, fmt clean. Independently reviewed (APPROVE-with-nits).

## Follow-ups (out of scope, not blocking)

1. **Expand a Codex `namespace` group into its constituent function sub-tools** when routing to Chat Completions, rather than dropping it — its sub-tools (`close_agent`, `send_input`, …) *do* have valid `{type:"function"}` forms, so expanding preserves Codex multi-agent capability. Needs response-side tool-call routing for the flattened names.
2. (Pre-existing) A `tool_choice` that names a now-dropped server tool could leave a dangling reference; consider normalizing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.0.0-alpha.14](https://github.com/bitrouter/bitrouter/compare/v1.0.0-alpha.13...v1.0.0-alpha.14)
+
+
+### 🐛 Bug Fixes
+
+- *(chat)* Drop provider-defined tools (no chat wire form) ([#553](https://github.com/bitrouter/bitrouter/pull/553)) - ([b0b4ef3](https://github.com/bitrouter/bitrouter/commit/b0b4ef365d83d9a5f2ee380067320ff5303d9f2c))
+
+
 ## [1.0.0-alpha.13](https://github.com/bitrouter/bitrouter/compare/v1.0.0-alpha.12...v1.0.0-alpha.13)
 
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -719,7 +719,7 @@ dependencies = [
 
 [[package]]
 name = "bitrouter"
-version = "1.0.0-alpha.13"
+version = "1.0.0-alpha.14"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -758,7 +758,7 @@ dependencies = [
 
 [[package]]
 name = "bitrouter-bedrock"
-version = "1.0.0-alpha.13"
+version = "1.0.0-alpha.14"
 dependencies = [
  "async-stream",
  "async-trait",
@@ -772,7 +772,7 @@ dependencies = [
 
 [[package]]
 name = "bitrouter-cloud-sdk"
-version = "1.0.0-alpha.13"
+version = "1.0.0-alpha.14"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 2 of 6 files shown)


FrozenArcher404:
⚠️ **Bundled emergency release.** This PR now also includes a `chore(release): publish v1.0.0-alpha.14` commit (workspace version + internal pins + `apps/bitrouter` mcp pin + `Cargo.lock` + `CHANGELOG.md`), done manually to skip the release-plz release-PR wait.

**Merging to `main` publishes alpha.14 automatically** — the `release-plz release` job runs on every push to main and publishes any crate whose manifest version isn't on crates.io yet, then tags `v1.0.0-alpha.14` (which triggers the cargo-dist `release.yml`). No manual `cargo publish` needed.

Caveat: the repo squash-merges, so the CHANGELOG commit link (`b0b4ef3`) points at the pre-merge fix commit; the `#553` PR link is canonical.

PR #569, opened 2026-06-15
chore(release): bump workspace to v1.0.0-alpha.16

FrozenArcher404:
Bumps the workspace version `1.0.0-alpha.15` → `1.0.0-alpha.16` (all crates inherit via `version.workspace`, plus internal dependency pins in the root `Cargo.toml`, the `bitrouter-mcp` pin in `apps/bitrouter/Cargo.toml`, and `Cargo.lock`).

This accompanies the alpha.16 release. The new leaf crate `bitrouter-attestation` was published to crates.io manually at `v1.0.0-alpha.16` to claim the name (the automated release token cannot publish brand-new crates); subsequent versions can go through trusted publishing.

Note: `bitrouter-attestation-plugin` is also a new, never-published crate and will need the same treatment (it can only be published after `bitrouter-sdk v1.0.0-alpha.16` is on crates.io, since it depends on it).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -743,7 +743,7 @@ dependencies = [
 
 [[package]]
 name = "bitrouter"
-version = "1.0.0-alpha.15"
+version = "1.0.0-alpha.16"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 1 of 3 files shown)


FrozenArcher404:
Superseded: we're dropping the `bitrouter-attestation` / `bitrouter-attestation-plugin` crates from mainline rather than releasing them (TEE verification will get a more careful, vendor-neutral design later). No standalone alpha.16 bump is needed; versioning returns to the normal release flow.

PR #581, opened 2026-06-18
fix(schema): don't embed workspace version in committed schema $id

FrozenArcher404:
## Root cause

`render_schema()` in `xtask/src/main.rs` stamped `CARGO_PKG_VERSION` into the schema's `$id` field:

```
"$id": "https://bitrouter.dev/schema/v1.0.0-alpha.15/config.schema.json"
```

When release-plz bumps the version to `alpha.16`, `cargo xtask generate-schema --check` regenerates the schema with `alpha.16` in `$id` — but the committed file still says `alpha.15`. String comparison fails, so the schema CI job fails on every release PR even though the schema structs haven't changed.

## Fix

Strip `$id` from the committed snapshot. The committed schema is a structural contract (do the serde structs match the file?), not a versioned artifact. A versioned URL belongs in the published/distributed schema; inject it at publish time, not at commit time. The `title` field is version-independent and is kept.

Regenerate the snapshot to drop the one `$id` line.

### DIFF
--- a/release-plz.toml
+++ b/release-plz.toml
@@ -10,6 +10,7 @@ pr_labels = ["release"]           # add the `release` label to the release Pull 
 [[package]]
 name = "xtask"
 release = false
+publish = false
 
 # Only the main binary package gets a tag, release, and changelog entry.
 [[package]]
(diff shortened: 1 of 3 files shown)


FrozenArcher404:
Superseded by #582 — adding a workflow step to regenerate the schema keeps the versioned $id while fixing the release PR CI failure.

Issue #390, opened 2026-04-24
feat: replace proxy takeover with wrapper shim + bitrouter claude CLI for agent launching


SPIKESPIGEL404:
### Use case

<h4 class="text-text-100 mt-2 -mb-1 text-base font-bold">Why this is better than config mutation</h4>
<div class="overflow-x-auto w-full px-2 mb-6">
  | Config mutation (cc-switch model) | Shim + CLI
-- | -- | --
~/.claude/config.json modified | ✅ yes | ❌ never
Crash recovery needed | ✅ yes | ❌ no
DB / backup table needed | ✅ yes | ❌ no
BitRouter down → claude broken | ✅ yes | ❌ falls back gracefully
Works from any terminal | ✅ yes | ✅ yes (via shim)

</div>

### Why is the current behavior insufficient?

The current approach to routing local Claude CLI traffic through BitRouter requires modifying ~/.claude/config.json directly (similar to cc-switch's start_with_takeover() model). This is invasive: it requires backup/restore logic, crash recovery, and leaves config in a broken state if the process dies mid-takeover.

### Proposed behavior

Replace the config-mutation approach with two complementary mechanisms:
1. bitrouter claude CLI command (spawn-time env override)
When BitRouter spawns a Claude Code agent, pass ANTHROPIC_BASE_URL=http://localhost:8787 directly in the child process environment via tokio::process::Command. ~/.claude/config.json is never touched. This already partially exists in bitrouter-providers/src/acp/provider.rs::resolve_launch.
2. Wrapper shim (optional user opt-in)
For users who want all claude terminal invocations to route through BitRouter without using the bitrouter claude command, BitRouter can install a shim at ~/.local/bin/claude:
bash#!/bin/bash
if curl -s http://localhost:8787/health > /dev/null 2>&1; then
    export ANTHROPIC_BASE_URL=http://localhost:8787
fi
exec /usr/local/bin/claude "$@"
The shim checks if BitRouter is running before setting the env var, so if BitRouter is down, claude falls back to hitting Anthropic directly — no broken state, no restore needed.

PR #584, opened 2026-06-18
feat(spawn): launch coding agents through BitRouter via `bitrouter spawn`

FrozenArcher404:
## Summary

Adds a `bitrouter spawn` subcommand that launches a coding-agent harness (Claude Code) as a child process with its API base URL pointed at the local BitRouter daemon — **without ever mutating the agent's own config files**. Routing lives only in the child's environment, so there is nothing to back up, restore, or recover after a crash, and if BitRouter is down the user simply runs the agent directly.

This implements the spawn-time env-override mechanism proposed in #390, choosing the explicit `spawn` verb (cargo-style) over a bare `bitrouter claude …` passthrough.

## CLI

Follows `cargo run`'s separator convention so bitrouter flags and agent flags are never ambiguous:

```
bitrouter spawn --agent/-a claude [--config <path>] [--base-url <url>] [--no-install] -- <args forwarded to claude>
```

Everything after `--` is forwarded to the agent verbatim, e.g.:

```
bitrouter spawn -a claude -- -p "summarize this repo" --dangerously-skip-permissions
```

## Behavior

- **Env-only routing.** Injects `ANTHROPIC_BASE_URL` (derived from `server.listen`; wildcard binds like `0.0.0.0`/`[::]` are rewritten to loopback, overridable with `--base-url`) into the child process only.
- **Gateway auth.** Injects `ANTHROPIC_AUTH_TOKEN` (sent as `Authorization: Bearer`, which is BitRouter's inbound scheme), not `ANTHROPIC_API_KEY` (which Claude Code sends as `x-api-key`). This is the documented way to route Claude Code through a custom LLM gateway. Token precedence: a token the user already exported → `BITROUTER_API_KEY` (`brk_…`) → a local placeholder that works under the `skip_auth` default. The auth token is always set explicitly, so an inherited `ANTHROPIC_API_KEY` (the *upstream* Anthropic provider key in this repo) can never silently become the inbound credential.
- **Install on demand.** When the agent binary is missing, offers to install it via the official native installer, gated by conditional compilation so only the host's method is ever shown — `curl -fsSL https://claude.ai/install.sh | bash` (macOS/Linux) or `irm https://claude.ai/install.ps1 | iex` (Windows). Non-interactive invocations (or `--no-install`) fail with the install command rather than hanging on a prompt.
- **Reachability hint.** Best-effort TCP probe warns (never blocks) when no daemon is listening.

## Extensibility

v1 ships Claude Code only. The `SpawnAgent` enum + `AgentSpec` metadata are the extension point — codex/gemini slot in without touching the dispatch, env-injection, or install machinery.

## Testing

- 11 new unit tests covering base-URL derivation (wildcard rewrite, bracketed IPv6, missing port), auth-token precedence, the never-inherit-`ANTHROPIC_API_KEY` guarantee, PATH resolution incl. the `~/.local/bin` post-install fallback, and the installer identity.
- `cargo test -p bitrouter --lib`: 135 passed, 0 failed.
- `cargo clippy -p bitrouter --bin bitrouter --lib`: clean.
- `#![forbid(unsafe_code)]` intact; PATH resolution is pure-std.

Third-party integration points (Claude Code env vars, install URLs) cite the official Claude Code docs in comments near the code.

Refs #390

### DIFF
--- a/apps/bitrouter/src/lib.rs
+++ b/apps/bitrouter/src/lib.rs
@@ -24,6 +24,7 @@ pub mod paths;
 pub mod policy;
 pub mod reload;
 pub mod skills;
+pub mod spawn;
 pub mod style;
 pub mod tools;
 
--- a/apps/bitrouter/src/main.rs
+++ b/apps/bitrouter/src/main.rs
@@ -3,7 +3,7 @@
 //! Subcommand surface: `serve` / `start` / `stop` / `restart` /
 //! `reload` / `status` / `route` / `init` / `key sign` / `models` / `tools` /
 //! `policy create` / `providers (list|login|logout)` / `agents` /
-//! `agent-proxy` / `cloud` / `skills`. Cloud-account sign-in lives under
+//! `agent-proxy` / `spawn` / `cloud` / `skills`. Cloud-account sign-in lives under
 //! `cloud (login|logout|whoami)`; per-provider credentials under
 //! `providers (login|logout)`. Daemon control runs over a local IPC endpoint
 //! (a Unix domain socket, or a Windows named pipe) — `start` spawns `serve`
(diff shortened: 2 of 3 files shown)


SPIKESPIGEL404:
@claude fix and pass fmt

PR #603, opened 2026-06-23
feat(server-tools): built-in web_search with BYOK backends

SPIKESPIGEL404:
## Summary

Adds a built-in `web_search` **router-executed server tool** with bring-your-own-key search backends. It follows the same pattern as advisor / sub-agent / fusion: enabled via `server_tools.web_search` config and advertised per request only when the caller declares `{"type":"bitrouter:web_search"}`. The result is that **any model routed through BitRouter** — even one with no native web search — gains a web search.

## Design

A `WebSearchBackend` trait is the single executor seam, with two implementors covering the four chosen engines plus a native-passthrough path:

- **`HttpSearchBackend`** — BYOK REST engines **Parallel / Exa / Firecrawl**. Keys resolve from an explicit `api_key` (supports `${VAR}`) or the conventional `*_API_KEY` env var; a backend with no resolvable key is skipped. Responses are parsed defensively through `serde_json::Value` so a provider renaming a field yields `None` rather than a hard failure.
- **`NestedSearchBackend`** — model-backed engines over the shared nested-completion runner: **Perplexity** (e.g. `perplexity/sonar`) and a provider's **native** web search (e.g. `anthropic:web_search_20250305`) **reused for every model**. This first cut is "option A": returns the synthesized `answer` with an empty `results` list (the answer embeds its sources).

**Backend selection mirrors BitRouter's provider model:** an ordered preference list (default = first), a per-request override that pins one backend by name, and **failover** to the next backend on error.

**Result schema** is deliberately minimal and stable:

```
{ backend, answer?, results: [ { url, title?, snippet?, content?, published?, score? } ] }
```

Every per-result field is optional (no engine fills them all), so the contract is additive — future backends extend via the trait, not the struct.

## Config

```yaml
server_tools:
  web_search:
    max_results: 5
    backends:                  # preference + failover order
      - kind: parallel         # PARALLEL_API_KEY
      - kind: exa              # EXA_API_KEY
      - kind: firecrawl        # FIRECRAWL_API_KEY
      - kind: perplexity       # nested; defaults to perplexity/sonar
      - kind: native           # reuse a provider's native search for all models
        model: anthropic/claude-opus-4.8
        tool: { type: "anthropic:web_search_20250305" }
```

## Changes

- New `server_tools/web_search/` module (`backend`, `http`, `nested`, `toolset`, `config`).
- `web_search` declaration parsed alongside advisor / sub-agent / fusion; `WebSearchSettings` added to `ServerToolsConfig`; wired in `assemble.rs::build_server_tool_loop` with key resolution + failover.
- Regenerated `schemas/bitrouter.config.schema.json`; documented in `skills/bitrouter/references/providers.md`.

## Tests / checks

21 new tests (request/response mapping per engine, declaration parsing, backend selection + failover, wiring). `cargo test --all-features`, `cargo clippy --all-features`, `cargo fmt --check`, and `xtask generate-schema --check` all pass.

## Notes for reviewers

1. HTTP request/response mappings are coded to the researched 2026 API shapes — most likely to need a tweak against a live key.
2. Nested backends ship as **option A** (answer text, no structured `results[]`); surfacing citations is a follow-up that needs a small `NestedOutcome` extension.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CCu2nEjnPimbY7atsU6dQ6)_

### DIFF
--- a/crates/bitrouter-sdk/src/language_model/server_tools/declarations.rs
+++ b/crates/bitrouter-sdk/src/language_model/server_tools/declarations.rs
@@ -28,6 +28,8 @@ use crate::plugin::PluginId;
 pub const ADVISOR_TOOL: &str = "advisor";
 /// Router-tool name the model calls to delegate a task to a worker model.
 pub const SUBAGENT_TOOL: &str = "subagent";
+/// Router-tool name the model calls to search the web.
+pub const WEB_SEARCH_TOOL: &str = "web_search";
 
 /// Plugin id under which [`ServerToolDeclarations`] is stashed on the request
 /// context by the pre-request hook, for the toolsets to read back.
@@ -48,6 +50,18 @@ pub struct AdvisorConfig {
     /// form; forwarded to the nested completion (see [`forwarded_tools`]).
     #[serde(default)]
     pub tools: Vec<serde_json::Value>,
+}
+
+/// Per-request Web-search config. Both fields are optional overrides: `backend`
+/// pins one of the configured search backends by name (else the default), and
+/// `max_results` lowers the per-call result cap. The query itself rides the
+/// tool call's arguments, not the declaration.
+#[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
+pub struct WebSearchDeclaration {
+    /// Pin a configured backend by name (else the deployment default).
+    pub backend: Option<String>,
+    /// Cap on results for this request (else the deployment default).
+    pub max_results: Option<u32>,
 }
 
 /// Per-request Sub-agent config (worker model + instructions + tools).
@@ -72,6 +86,8 @@ pub struct ServerToolDeclarations {
     pub subagent: Option<SubAgentConfig>,
     /// The fusion declaration, if any (already resolved against `parent_model`).
     pub fusion: Option<FusionConfig>,
+    /// The web-search declaration, if any.
+    pub web_search: Option<WebSearchDeclaration>,
     /// The outer request model — the default nested model.
     pub parent_model: String,
 }
@@ -108,6 +124,12 @@ impl ServerToolDeclarations {
                         tools: array_field(args, "tools"),
                     });
                 }
+                Some(Kind::WebSearch) => {
+                    decls.web_search = Some(WebSearchDeclaration {
+                        backend: str_field(args, "backend"),
+                        max_results: u32_field(args, "max_results"),
+                    });
+                }
                 None => {}
             }
         }
@@ -123,21 +145,26 @@ impl ServerToolDeclarations {
 
     /// Whether the request declared no server tool.
     pub fn is_empty(&self) -> bool {
-        self.advisor.is_none() && self.subagent.is_none() && self.fusion.is_none()
+        self.advisor.is_none()
+            && self.subagent.is_none()
+            && self.fusion.is_none()
+            && self.web_search.is_none()
     }
 }
 
 enum Kind {
     Advisor,
     SubAgent,
+    WebSearch,
 }
 
-/// Recognise an advisor / sub-agent declaration by tool name: the bare name or a
-/// namespaced form whose final `:`/`.` segment is the name.
+/// Recognise an advisor / sub-agent / web-search declaration by tool name: the
+/// bare name or a namespaced form whose final `:`/`.` segment is the name.
 fn server_tool_kind(name: &str) -> Option<Kind> {
     match name.rsplit([':', '.']).next().unwrap_or(name) {
         ADVISOR_TOOL => Some(Kind::Advisor),
         SUBAGENT_TOOL => Some(Kind::SubAgent),
+        WEB_SEARCH_TOOL => Some(Kind::WebSearch),
         _ => None,
     }
 }
@@ -149,6 +176,15 @@ fn str_field(args: &serde_json::Value, key: &str) -> Option<String> {
         .or_else(|| args.get("parameters").and_then(|p| p.get(key)))
         .and_then(|v| v.as_str())
         .map(str::to_string)
+}
+
+/// Read a `u32` field from a config object (same `parameters`-wrapper tolerance
+/// as [`str_field`]).
+fn u32_field(args: &serde_json::Value, key: &str) -> Option<u32> {
+    args.get(key)
+        .or_else(|| args.get("parameters").and_then(|p| p.get(key)))
+        .and_then(|v| v.as_u64())
+        .map(|n| n.min(u32::MAX as u64) as u32)
 }
 
 /// Read an array field from a config object (same `parameters`-wrapper tolerance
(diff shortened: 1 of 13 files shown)


crates/bitrouter-sdk/src/language_model/server_tools/declarations.rs
```diff
@@ -123,21 +145,26 @@ impl ServerToolDeclarations {
...
 
-/// Recognise an advisor / sub-agent declaration by tool name: the bare name or a
-/// namespaced form whose final `:`/`.` segment is the name.
+/// Recognise an advisor / sub-agent / web-search declaration by tool name: the
+/// bare name or a namespaced form whose final `:`/`.` segment is the name.
 fn server_tool_kind(name: &str) -> Option<Kind> {
     match name.rsplit([':', '.']).next().unwrap_or(name) {
         ADVISOR_TOOL => Some(Kind::Advisor),
         SUBAGENT_TOOL => Some(Kind::SubAgent),
+        WEB_SEARCH_TOOL => Some(Kind::WebSearch),
```
FrozenArcher404: `server_tool_kind` matches on the bare tail name, and `from_prompt` only inspects `ProviderDefined` tools — so **any** provider tool whose tail is `web_search` sets `decls.web_search`, including a caller's *native* search tool (e.g. OpenAI's Responses `web_search`). For `advisor`/`subagent`/`fusion` the bare-tail match is safe (no upstream uses those names), but `web_search` genuinely collides with real provider tools.

Consequence when `server_tools.web_search` is enabled: a request carrying a native `web_search` tool is parsed as a `bitrouter:web_search` declaration, the toolset advertises, and `inject` then strips the caller's native tool (its tail is now in `owned`) — silently swapping the model's native search for the BYOK backend.

This also makes the comment at `loop_controller.rs:83-86` inaccurate: *"a caller's native `web_search` when `bitrouter:web_search` was not declared is left untouched"* can't happen for a tool literally named `web_search`, because its presence **is** the declaration. The preservation only protects differently-named native tools (Anthropic's `web_search_20250305`, etc.), which is what the `tool_tail` doc example correctly shows.

Options: (a) keep the behavior but fix the loop_controller comment to use `web_search_20250305` as the example and document that a native `web_search` gets taken over; or (b) require the explicit `bitrouter:` namespace for this declaration, so a bare/foreign-namespaced `web_search` is left to the upstream unless the caller opts in. (Worth double-checking the exact upstream `type` strings.)

crates/bitrouter-sdk/src/language_model/server_tools/web_search/toolset.rs
```diff
@@ -0,0 +1,319 @@
...
+            return Ok(error_output("web_search call is missing required `query`"));
+        };
+        // Result cap precedence: call arg → declaration → deployment default.
+        let max_results = args
+            .get("max_results")
+            .and_then(|v| v.as_u64())
+            .map(|n| n.min(u32::MAX as u64) as u32)
+            .or(decl.max_results)
+            .unwrap_or(self.default_max_results)
+            .max(1);
```
FrozenArcher404: The call-arg `max_results` overrides the declaration cap and the deployment `default_max_results` with no upper bound (`.or(decl.max_results).unwrap_or(self.default_max_results).max(1)`), so the model can request **more** than configured. That contradicts the documented intent — `config.rs` calls it a "maximum ... (caller may lower it)" and the declaration field "lowers the per-call result cap" — and the advertised tool schema sets `minimum: 1` with no maximum. HTTP backends pass this straight through as `numResults` / `limit` / `max_results`, so `max_results: 100` inflates response size and cost past the operator's (and client's) cap.

If these are meant as ceilings, clamp the call-arg, e.g. `min(requested, decl.or(default))`. If they're really just defaults the model may raise, reword the config/declaration docs to drop the "cap" / "may lower it" framing.

crates/bitrouter-sdk/src/language_model/server_tools/loop_controller.rs
```diff
@@ -247,6 +251,14 @@ impl ServerToolLoop {
 
 /// Append the model's tool-call turn and the tool-result turn to the working
 /// prompt so the next upstream call sees the results.
+/// The bare tail of a (possibly namespaced) tool name: the final `:`/`.`
+/// segment. Matches a `bitrouter:fusion` declaration against the advertised
+/// bare `fusion`, while leaving an unrelated provider tail (e.g.
+/// `web_search_20250305`) distinct from `web_search`.
+fn tool_tail(name: &str) -> &str {
```
FrozenArcher404, line 256: The doc comment got split here: lines 252-253 (*"Append the model's tool-call turn ..."*) document `append_turn`, but inserting `tool_tail` between them and `append_turn` leaves them stranded — `tool_tail` now carries a 6-line doc whose first two lines describe a different function, and `append_turn` (line 262) is left undocumented. Move `tool_tail` and its own 4-line doc out from between them (above this pair, or below `append_turn`) so each function keeps its doc.

crates/bitrouter-sdk/src/language_model/server_tools/web_search/toolset.rs
```diff
@@ -0,0 +1,319 @@
...
+        if let Some(name) = override_name
+            && let Some(pinned) = self.backends.iter().find(|b| b.name() == name)
+        {
+            ordered.push(pinned);
+        }
+        for b in &self.backends {
+            if !ordered.iter().any(|o| o.name() == b.name()) {
+                ordered.push(b);
+            }
+        }
```
FrozenArcher404: `candidates` dedups by `name()`, so two backends that share a name collapse to the first and the rest become unreachable via failover or a `backend` override. Easy to hit: the `perplexity` kind has no `name` field (two perplexity backends always collapse), and two `native` backends without an explicit `name` both default to `"native"`. Since the feature is an ordered failover list, consider deduping by identity/index (skip only the specific pinned entry) rather than by name — or validate that backend names are unique.

FrozenArcher404 (commented):
Reviewed the `web_search` server-tool PR. Overall it's clean and consistent with the existing advisor / sub-agent / fusion server tools: the `WebSearchBackend` seam, the ordered preference/failover list, the defensive `serde_json::Value` response parsing, the per-request advertisement gating, and the `inject` change (strip only *advertised* tools rather than anything the registry could own) are all well done, and test coverage is solid.

A few things worth a look before merge — comments inline. The main one is the `web_search` name collision with providers' native tools (declarations.rs); the rest are a cost-control gap on `max_results`, a stranded doc comment, and a backend-dedup edge case. I left the HTTP request/response field mappings alone since you already flagged them for live-key verification.

FrozenArcher404:
Addressed these review points in #606, which targets this branch (`claude/web-search-server-tools-74j5li`):

- **`web_search` name collision** — gated the declaration on the explicit `bitrouter` namespace, so a caller's native `web_search` (e.g. OpenAI's Responses tool) is no longer parsed as a declaration and stripped. This also makes the `loop_controller` "left untouched when undeclared" comment accurate.
- **`max_results`** — clamped so each layer can only lower the cap (deployment default is the hard ceiling).
- **`candidates` dedup** — now by index, so same-named backends stay in the failover chain.
- **stranded doc** — `append_turn`'s doc restored.

Each has a test; CI is green on #606.

PR #606, opened 2026-06-24
fix(server-tools): address web_search review follow-ups

FrozenArcher404:
## Summary

Follow-ups to the built-in `web_search` server tool, addressing the inline review on this PR. Targets `claude/web-search-server-tools-74j5li` so it merges **into** the feature branch (not `main`).

## Fixes

1. **Gate the `web_search` declaration on the `bitrouter` namespace.** Unlike the bitrouter-invented `advisor` / `subagent` / `fusion` names, `web_search` is also a real *native* tool name on several providers (e.g. OpenAI's Responses `web_search`). `server_tool_kind` matched the bare tail, so a caller's native `web_search` was parsed as a `bitrouter:web_search` declaration, advertised, and then **stripped from the upstream prompt** by the loop's `inject` step — silently replacing the model's native search with the BYOK backend. Now only `{"type":"bitrouter:web_search"}` declares the built-in tool; a bare or foreign-namespaced `web_search` is left for the upstream. This also makes the existing `loop_controller` comment ("a caller's native `web_search` … is left untouched") actually true.

2. **Clamp `max_results` to the configured ceiling.** The per-call argument could exceed both the per-request declaration cap and the deployment `default_max_results`, despite the docs describing those as caps the caller "may lower". Now each layer can only *lower* the cap: deployment default is the hard ceiling ≥ declaration ≥ call argument. Prevents a model from inflating result count / cost past what the operator allowed.

3. **Keep same-named backends in the failover chain.** `candidates` deduped by `name()`, so two backends sharing a name (e.g. two `native` backends left at the default name, or two `perplexity`) collapsed to the first and the rest were unreachable via failover/override. Now the pinned backend is excluded by **index**, not name.

4. **Restore `append_turn`'s doc comment**, which had become attached to `tool_tail` when that helper was inserted.

## Tests / checks

Adds tests for each fix (namespace gating accepts `bitrouter:web_search` and rejects bare / foreign `web_search`; `max_results` clamps to the ceiling; same-named backends both stay reachable on failover). Locally green:

- `cargo fmt --all -- --check`
- `cargo clippy --workspace --all-features --tests -- -D warnings`
- `cargo nextest run --workspace --all-features` (1169 passed)
- `cargo xtask generate-schema --check` (schema unchanged — no config struct changes)
- `cargo test --doc`

No config-schema or public-API changes.

### DIFF
--- a/crates/bitrouter-sdk/src/language_model/server_tools/declarations.rs
+++ b/crates/bitrouter-sdk/src/language_model/server_tools/declarations.rs
@@ -160,13 +160,31 @@ enum Kind {
 
 /// Recognise an advisor / sub-agent / web-search declaration by tool name: the
 /// bare name or a namespaced form whose final `:`/`.` segment is the name.
+///
+/// `web_search` is the exception. Unlike the bitrouter-invented
+/// `advisor` / `subagent` / `fusion` names, `web_search` is also a real
+/// *native* tool name on several providers (e.g. OpenAI's Responses
+/// `web_search`). Matching it bare would hijack a caller's genuine native tool,
+/// so it is recognised only under the explicit `bitrouter` namespace
+/// (`{"type":"bitrouter:web_search"}`, the documented declaration form); a bare
+/// or foreign-namespaced `web_search` is left untouched for the upstream.
 fn server_tool_kind(name: &str) -> Option<Kind> {
     match name.rsplit([':', '.']).next().unwrap_or(name) {
         ADVISOR_TOOL => Some(Kind::Advisor),
         SUBAGENT_TOOL => Some(Kind::SubAgent),
-        WEB_SEARCH_TOOL => Some(Kind::WebSearch),
+        WEB_SEARCH_TOOL if is_bitrouter_namespaced(name) => Some(Kind::WebSearch),
         _ => None,
     }
+}
+
+/// Whether `name` carries the explicit `bitrouter:` / `bitrouter.` namespace —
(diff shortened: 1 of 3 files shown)


FrozenArcher404:
This PR implements the four follow-ups raised in the inline review of #603:

| Review finding | Fix in this PR |
| --- | --- |
| `web_search` bare-tail match hijacked a caller's **native** `web_search` (the `loop_controller` "left untouched" comment was unreachable) | `server_tool_kind` now gates `web_search` on the explicit `bitrouter` namespace — verified across the Messages / Responses / GenerateContent decoders that the `bitrouter:` prefix survives in the canonical tool `name` (Chat Completions can't carry a provider-defined declaration at all). The comment is now actually true. |
| `max_results` call-arg could exceed the configured cap despite "caller may lower it" | Layered clamp: deployment `default_max_results` ≥ declaration ≥ call-arg, each only lowering. |
| `candidates` deduped by name → same-named backends collapsed, breaking failover/override | Dedup by **index** — the pinned backend is excluded by position, so same-named backends stay in the failover chain. |
| `append_turn`'s doc comment stranded on `tool_tail` | Reordered so each function keeps its own doc. |

Each fix has a dedicated test. Local `fmt` / `clippy -D warnings` / `nextest` (1169 passed) / `generate-schema --check` / doc-tests are all green; no config-schema or public-API change.

Chose the namespace-gating option (over just rewording the comment) because it matches this PR's own stated contract — "advertised only when the caller declares `{"type":"bitrouter:web_search"}`" — and keeps a caller's genuine native search intact.

Issue #249, opened 2026-04-04
Delete .claude/skills/a2a.md?


FrozenArcher404:
@SPIKESPIGEL404 Delete this?

PR #613, opened 2026-06-28
feat(acp): per-session ACP substrate (acp serve|prompt)

SPIKESPIGEL404:
## Summary

Replaces the opaque 1:1 `bitrouter agent-proxy` with a **per-session ACP substrate** — one process = one session = one upstream agent — exposed as the primary headless CLI and as a vanilla ACP endpoint any *manager* (the GUI, an editor, or an AI manager-agent) can drive. Orchestration lives in the manager, not the substrate.

- **New crate `bitrouter-substrate`** — owns one session end-to-end: upstream `agent-client-protocol` `Client` connection (callback plane for streaming + permissions), a single-writer FIFO turn queue, git-worktree lifecycle, three-tier session identity, and bitrouter policy via the retained `acp::Pipeline` + a telemetry `ExecutionHook`.
- **New CLI (primary surface):**
  - `bitrouter acp serve --agent <id> [--worktree <name>]` — serves one session as a vanilla ACP Agent over stdio (managers spawn one per session).
  - `bitrouter acp prompt --agent <id> [--worktree <name>] [--no-wait] <text>` — headless one-shot, streams NDJSON, exits.
- **SDK `acp` core retyped** over runtime-free `agent-client-protocol-schema` (typed `PromptRequest`/`PromptResponse`, zero round-trip).
- **Removed:** `agent-proxy` (CLI + `agent_proxy.rs`), `AcpStdioExecutor`, and the dead app-level `acp::Pipeline` wiring. `bitrouter agents check` migrated to a typed `initialize` health-check (with agent `env` correctly propagated).

### Design (decisions D1–D13 in the spec)
Per-session, no central daemon / no `_conductor/*` (vanilla ACP; agent + worktree are launch args); one agent per session (no mid-session switch); FIFO turn queue with cancellation at the upstream level; three-tier identity (`record_id` / `acp_session_id` / `agent_session_id`); ephemeral v1 sessions (warm per-session owners are v2).

### v1 limitations (documented follow-ups)
- `fs/*` / `terminal/*` answered `method_not_found` (sandboxed handlers TBD).
- Permission outcomes are coarse (`allow`/`deny`; exact `optionId` not preserved).
- Telemetry is `{agent, stop_reason}` (token/latency enrichment via `UsageUpdate` + a `PreRequest` timestamp hook is a follow-up).
- **Cross-repo follow-up:** repoint `bitrouter-gui`'s `AcpFeed` from `agent-proxy` to `bitrouter acp serve --agent <id>` (gated on this landing).

## Test Plan
- [x] `cargo nextest run --all-features` — **1219 passed, 2 skipped**
- [x] `cargo clippy --all-features` — clean
- [x] Real subprocess E2E (`serve`) over OS stdio; in-memory duplex E2E covering `session/new` → `prompt` → `session/update`, the **permission round-trip**, `session/cancel`, and `method_not_found`
- [x] Manual smoke: `bitrouter acp prompt --agent claude-acp "say hi"` against a configured agent


SPIKESPIGEL404:
### Follow-ups tracked (deferred from this PR)

v1 limitations / v2 work, each filed as its own issue:

- #614 — sandboxed `fs/*` + `terminal/*` handlers for `acp serve` (R1; currently `method_not_found`)
- #615 — preserve exact permission `optionId` end-to-end (R5; outcomes are coarse today)
- #616 — enrich turn telemetry with tokens + latency (currently `{agent, stop_reason}` only)
- #617 — forward non-text prompt content blocks (text-only today)
- #618 — **v2:** warm/persistent sessions — owners, on-disk persistence, resume (sessions are ephemeral today; unlocks `--session` + multi-turn `prompt`)
- #619 — **v2:** fan-out/broadcast + spawn-time routing (one agent per session today)

Cross-repo: bitrouter/bitrouter-gui#3 — repoint `AcpFeed` from `agent-proxy` to `acp serve` (gated on this PR).

SPIKESPIGEL404:
## Refinement round: proxy fidelity + session operability

This push extends the substrate with two rounds of refinements on top of the original v1. First, **wire fidelity** — closing every gap where the proxy silently degraded what real agents/managers do:

- **Exact permission `optionId` pass-through** (6562f02a) — the manager's selection previously collapsed to a coarse allow/deny enum and was re-guessed by option *kind* upstream, so two same-kind options ("allow this command" vs "allow all npm commands") were indistinguishable. The resolver now carries the exact `RequestPermissionOutcome`, validated against the offered set; drop still defaults to the reject option.
- **Multi-modal prompts forward verbatim** (7d9ac8d0) — `session/prompt` content blocks (images, resources, resource links) reach the agent unmodified instead of being text-flattened.
- **Capability relay** (b1c59a20) — downstream `initialize` reflects the upstream agent's real capabilities + `agentInfo` instead of a fabricated minimal set.

Second, **operability**:

- **Worktrees retained by default** (53fc5409) — shutdown previously ran `git worktree remove --force` unconditionally, destroying the agent's uncommitted work (and `serve` never cleaned up at all). Retention is now the default with `--rm-worktree` opt-in; `WorktreeManager` reuses existing worktrees and attaches existing branches.
- **Deterministic shutdown** (875e739f) — the `Arc::strong_count` yield-polling heuristic is replaced by an explicit `Shutdown` command with teardown confirmation.
- **Honest telemetry + usage NDJSON** (94f62417) — the always-0 token placeholder fields are gone; per-turn records carry real `latency_ms` + context occupancy from `UsageUpdate`, which also streams as a `{"type":"usage",…}` NDJSON line. (ACP's stable surface has no per-turn token deltas — only context `used`/`size`.)
- **Durable session records + `acp sessions`** (aaab6d4e) — `.bitrouter/sessions/<record_id>.json` per session, with pid-liveness (`dead`) detection in the listing.
- **Durable transcript + session-scoped cancel + turn timeout** (4b4d650b) — a non-lossy NDJSON transcript per session (`{seq, ts, kind}` lines; `--no-transcript` opt-out); `session/cancel` now also flushes the queued backlog (matching the spec's session-scoped semantics); `--turn-timeout` with a cooperative-cancel grace.
- **ACP registry tier 1** (0820f657) — `agents list --remote` fetches the official registry (~38 agents); `agents install` falls back to it for `npx`/`uvx` distributions (version-pinned, env included). Binary-only entries are refused with a manual-install pointer — the registry publishes no checksums, so tier 2 stays deferred.

Test count went from 1219 → 1256 (all passing), clippy + fmt clean throughout.

SPIKESPIGEL404:
## ACP v2 alignment: handshake relay, session/load replay, warm sessions, agent spans

The second half of this push implements the v2-alignment sequence, designed after reading the [ACP v2 RFDs](https://agentclientprotocol.com/rfds/v2/overview). Two RFD findings shaped it:

1. **v2 removes the client `fs/*`/`terminal/*` surface** (low adoption; client-side MCP servers are the replacement) — so the planned fs/terminal pass-through plane was dropped from the roadmap entirely, replaced by relaying the manager's MCP servers.
2. **v2 replaces `session/load` with `session/resume { replayFrom }`** and standardizes remote transports (streamable HTTP / WebSocket, agent-as-server, sessions persisting across reconnects) — so warm sessions here are built to be *convergent*: same semantics, swappable transport.

**Phase (a) — deferred `session/new` with cwd + `mcpServers` relay** (ac7d9a8e). `serve` now spawns + initializes the agent up front (so capabilities reflect to the manager) but creates the upstream session only when the *manager's* `session/new` arrives, relaying its `cwd` (launch-time `--worktree` wins) and `mcpServers` verbatim. This is how a v2-aligned manager provides fs/terminal-style tooling. The headless `prompt` path keeps immediate-open behavior; `session/new` stays idempotent (same stable `record_id`).

**Phase (b) — `session/load` replays the transcript** (6be4f86a). v1 `session/load` with v2 `session/resume {replayFrom: start}` semantics: the durable transcript replays as `session/update` notifications (user prompts as `user_message_chunk`, upstream updates verbatim) before the response, then the session continues live. `loadSession` is now the substrate's own capability, advertised exactly when a transcript exists. The transcript's writer-minted monotonic `seq` is the future `replayFrom` cursor.

**Phase (c) — warm sessions** (03f251ce). `serve --warm [--idle-timeout SECS]` survives manager disconnects, accepting reattach on a per-session unix socket that speaks the **exact stdio NDJSON JSON-RPC framing** via the SDK's `ByteStreams` — deliberately no bespoke protocol, so ACP's standardized remote transport replaces the pipe without touching anything above it. `bitrouter acp attach <record>` bridges a terminal to the socket (pure byte pump). Covered by a subprocess e2e over the full lifecycle: stdio session → disconnect → socket reattach → `session/load` replay → live prompt → idle-timeout self-shutdown. Sockets bind under `$BITROUTER_HOME/sock`, not the repo — unix `sun_path` caps at ~104 bytes on macOS and repo paths blow through it (found the hard way).

**OTel GenAI agent spans** (8092533f). New `acp` module in `bitrouter-observe`: an INTERNAL `invoke_agent <agent>` span per turn (back-dated start so duration = real latency) and `execute_tool <tool>` spans per completed tool call, correlated by `gen_ai.conversation.id` = `record_id` — which is also the join key to the HTTP router plane when the fronted agent routes its model calls through bitrouter. Context occupancy rides `bitrouter.context.*` (deliberately not `gen_ai.usage.*` — those are per-turn token deltas ACP doesn't provide). Wired behind the same `plugins.bitrouter-observe` config as the daemon; off by default; a dead OTLP endpoint never blocks or breaks a session. The substrate crate stays OTel-free.

Also in this push: `.bitrouter/` gitignored + leaked e2e test artifacts untracked (75409470), and `skills/bitrouter/` docs updated throughout per the repo rule.

Full suite: **1256 passed / 0 failed**, clippy + fmt clean.

SPIKESPIGEL404:
## Fix from real-agent e2e: process-group teardown (no more orphaned agents)

Driving the CLI against the real `claude-code-acp` surfaced two teardown bugs, both now fixed:

1. **Headless permission hang** (b949f837) — a `session/request_permission` during `acp prompt` parked forever because headless mode never consumed the permission stream. Prompt now auto-denies (logged), and the turn completes.
2. **Orphaned agent processes** (6ea816a2) — the SDK's `ChildGuard` SIGKILLs only the immediate child, but the default catalog invocation is a wrapper chain (`npx → node`), so the *real* agent re-parented to pid 1 and lingered after every session teardown. The substrate now owns the spawn: agent children launch in their **own process group** (handed to the SDK via `ByteStreams`), and a reaper task group-kills + reaps on every exit path — explicit shutdown, agent self-exit, and kill-order sender drop (panics/timeouts collapse to the same cleanup). Replacing `AcpAgent` also revealed its child-monitor race was load-bearing (`ByteStreams` EOF doesn't fail in-flight requests), so the connection future is raced against the death signal — prompts to a crashed agent now fail fast instead of hanging.

Verified against real claude-code-acp: zero surviving processes after `acp prompt` and `agents check` through a 4-level `env → npx → npm → node` chain. Full e2e also exercised: registry install → `agents check` (559ms initialize), NDJSON prompt streaming, worktree retention/reuse, real capability relay (`image:true`, `agentInfo: Claude Code v0.16.2`), a manager-approved permission that actually wrote a file, `session/load` replay on reattach over the warm socket, and idle reap. Suite now at **1259 passed**.

SPIKESPIGEL404:
Upstreamed the orphaned-process-tree fix to the ACP Rust SDK (the ecosystem-wide half of 6ea816a2): [agentclientprotocol/rust-sdk#251](https://github.com/agentclientprotocol/rust-sdk/pull/251) (fixes [#249](https://github.com/agentclientprotocol/rust-sdk/issues/249)), plus [#250](https://github.com/agentclientprotocol/rust-sdk/issues/250) for the transport-EOF/in-flight-request semantics gap that made our child-death race necessary. bitrouter keeps its own spawn regardless (reaper exit-path unification + roadmap hooks); if upstream lands the group-kill, our belt-and-braces margins just get thinner.

PR #624, opened 2026-06-30
feat(routing): config-driven policy-table model router

FrozenArcher404:
## What

Adds a **config-driven per-request model router** to the daemon: a
`PolicyTableRouter` `PromptTransform` that rewrites `prompt.model` based on a
static policy table, so a long agent loop can spend a capable model only where
it matters and a cheaper one on routine steps — without the caller changing how
it addresses the router.

## How it works

- **Fingerprint** (deterministic, from the canonical `Prompt`, by the model's
  most recent turn): `opening` (no model turn yet), `after_<tool>` (the model's
  last turn called `<tool>`), or `midstream` (a model turn with no tool call).
- **Tier resolution**: fingerprint → tier (via `fingerprints`), falling back to
  `default_tier`; tier → model id (via `tiers`).
- **Tool-use guardrail**: a request carrying tools is clamped up to
  `tool_use_tier` unless its chosen tier is listed in `tool_safe_tiers`.
- **Rewrite**: `prompt.model` is set to the resolved model. No inference, no LLM.

It mirrors `ClaudeCodeRouter` and `FusionAliasConfig`. It is idempotent and
defers to routes it does not own: an explicit `provider:model` id (including
`claude-code:`) and any request carrying a bitrouter server-tool declaration
(e.g. the `bitrouter/fusion` alias's injected tool) are left untouched. It is
registered after the fusion alias and the Claude Code router so those win.

## Config

New top-level `policy_table:` section (see `examples/bitrouter.yaml`):

```yaml
policy_table:
  tiers:
    cheap: openai:gpt-4o
    capable: anthropic:claude-sonnet-4-6
  fingerprints:
    opening: capable
    after_read_file: cheap
  default_tier: capable
  tool_use_tier: capable
  tool_safe_tiers: [capable]
```

Inert unless `tiers` is non-empty. Validated at parse time: a fingerprint /
default / guardrail tier not defined in `tiers` is a 400, and a `tool_use_tier`
that isn't itself listed in `tool_safe_tiers` is a 400.

## Tests / checks

- Unit tests for the transform (fingerprint cases incl. the moved-past-a-tool
  `midstream` case and parallel tool calls, tier resolution, default tier,
  tool-use guardrail + its disabled form, idempotence on both bare and colon-form
  tier targets, explicit-route skip, fusion-declaration skip, unconfigured no-op)
  and config (defaults, round-trip, unknown-tier + guardrail-target rejection).
[...]

PR #625, opened 2026-06-30
feat(routing): online adequacy ledger for adaptive policy-table routing

FrozenArcher404:
**Stacked on #624** (review/merge that first; this targets `feat/policy-table-router`).

## What

Makes the policy-table router (#624) **adaptive**. A local, telemetry-fed **adequacy ledger** learns, per request fingerprint, when a downgraded tier keeps hard-failing and **escalates** that fingerprint to a more capable tier. This is the *safety half* of the asymmetric routing rule: an operator's downgrade that proves inadequate self-corrects, online, with no LLM in the loop. The learner can only ever make routing **more conservative** than the static table — it never downgrades on its own (that's the follow-up, "aggressive" half).

## How (credit assignment, no round structure)

An `ObserveHook` runs after every request and:
1. recomputes the fingerprint from the prompt,
2. maps the served model back to its tier,
3. credits the outcome **only to a genuine policy-router downgrade** — the served tier must equal the tier the *static* table resolves for this request (so a caller's explicit route, a coincidental model match, or an escalation is not mistaken for a downgrade) and must not be the escalation tier,
4. records whether it hard-failed (`Failed`; `ClientDisconnected` skipped; `Completed` = held).

After `escalation_threshold` consecutive failures the fingerprint is **pinned**; the ingress router routes it to the escalation tier. A pin decays after `pin_cooldown_secs` (the downgrade is retried later). Pins persist in the local db (`adequacy_pins`) and the in-memory cache the router reads on the hot path is warmed from it at startup.

The read path (`is_pinned`) is sync (no async in the `PromptTransform`); the write path (`observe`) is async (persists). The static `PolicyTable` spec is extracted behind an `Arc`, shared by the router (reads) and the observer (recompute + reverse model→tier map).

## Config (off by default)

```yaml
policy_table:
  tiers: { cheap: openai:gpt-4o, capable: anthropic:claude-sonnet-4-6 }
  fingerprints: { after_read_file: cheap }
  default_tier: capable
  adequacy:
    enabled: false          # opt-in
    escalation_tier: capable # defaults to default_tier
    escalation_threshold: 2  # consecutive hard failures before pinning
    pin_cooldown_secs: 1800  # 0 = no decay
```

Validated at parse time (unknown escalation tier; enabled-without-target). Inert unless `adequacy.enabled`.

## Incidental SDK fix

The pipeline now surfaces a streaming request's terminal outcome to `on_request_end`: a mid-stream `StreamOutcome::UpstreamError` maps to `RequestOutcome::Failed` (a `StreamHook` abort stays `Completed`), instead of reporting every streamed request as `Completed`. This is a latent observability gap any error-rate observe hook depends on — and the ledger needs it to learn from the dominant (streaming) request shape.

## Tests / checks

Ledger (threshold/pin/reset/cooldown, persistence round-trip + survives-restart); observer (genuine-downgrade gate, failure→pin, coincidental-match ignored, escalation-tier ignored, completed/disconnect ignored); adaptive router (pinned→escalate; #624 static behavior preserved when off); pipeline outcome mapping; config defaults + validation. `cargo nextest -p bitrouter-sdk -p bitrouter --all-features`: **801 pass**. clippy `-D`, fmt, `generate-schema --check`, rustdoc `-D`, `config validate` all clean.

### DIFF
--- a/apps/bitrouter/src/adequacy/mod.rs
+++ b/apps/bitrouter/src/adequacy/mod.rs
@@ -1,0 +1,203 @@
+//! Online adequacy ledger — the escalate-sticky learned state behind adaptive
+//! policy-table routing.
+//!
+//! When `policy_table.adequacy` is enabled, the daemon watches every request a
+//! *downgrade* produced (the policy table routed a fingerprint to a cheaper tier
+//! than its escalation tier) and counts the ones that hard-fail. Once a
+//! fingerprint accumulates `escalation_threshold` failures it is **pinned**: the
+//! adaptive [`crate::policy_table_router::PolicyTableRouter`] then routes that
+//! fingerprint to the escalation tier instead of the cheap one. A pin decays
+//! after `pin_cooldown_secs`, so a downgrade that failed transiently is retried
+//! later rather than abandoned forever.
+//!
+//! This is the safety half of the asymmetric routing rule: the ledger never
+//! downgrades on its own — it only escalates a downgrade that is failing — so it
+//! can only ever make routing *more* conservative than the static table.
+//!
+//! The read path ([`AdequacyLedger::is_pinned`]) is synchronous and lock-cheap
+//! because it runs inside the ingress [`bitrouter_sdk::PromptTransform`]; the
+//! write path ([`AdequacyLedger::observe`]) is async because it persists new
+//! pins, and runs from the [`observer::AdequacyObserveHook`] after each request.
+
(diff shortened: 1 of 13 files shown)


FrozenArcher404:
Consolidated into #627 — the entire feature (P1 + P2 + P3) in a single branch against `main`. Closing this stacked PR; the commit and its review history are preserved in #627.

PR #626, opened 2026-06-30
feat(routing): aggressive downgrade discovery + convergence eval

FrozenArcher404:
**Stacked on #625** (which is stacked on #624). Review/merge those first; this targets `feat/adequacy-ledger`.

## What

The *aggressive* half of the asymmetric routing rule, on top of the adequacy ledger's safety half (#625). When exploration is enabled, the daemon periodically **trials** the cheap `explore_tier` on fingerprints the static table routes to the escalation tier — ones the operator did **not** downgrade — and **locks** a fingerprint to the cheap tier once it survives `explore_threshold` adequate trials. Safe downgrades are discovered automatically rather than only hand-configured. A trial that fails escalates and stops (the same sticky pin as the safety path), so the guarantee holds: **never net-lose, while still pursuing cheaper routes.**

Deterministic (the trial cadence is a counter — no randomness), opt-in (`policy_table.adequacy.explore_enabled`), and the tool-use guardrail still clamps any trial that would downgrade a tool request below the floor.

## How

- **Router decision order** (`adapt`): explicit/server-tool guards → static decision → `pin?`→escalate → `locked?`→cheap → `trial due?`→cheap → guardrail. P2/#624 behavior is preserved exactly when exploration is off or no ledger is wired.
- **Ledger**: a unified per-fingerprint `Entry` (failure tally + pin + trial cadence counter + adequate-trial tally + lock). `should_trial` fires ~1/`explore_interval` deterministically. A failed trial un-locks and pins.
- **Observer**: classifies each request into a typed `Outcome` — `StaticDowngrade` (served == static decision, a downgrade) vs `Exploration` (static decision is the escalation tier; `trialed` = served the explore tier vs cadence-advance on the escalation tier). The two branches partition on `static == escalation`, so a trial can't be miscounted as a static downgrade or vice versa.
- **Persistence**: pins persist (safety); locks are in-memory and cheaply re-discovered.

## Convergence eval

`apps/bitrouter/src/adequacy/eval.rs` drives the **real** router + observer + ledger over a synthetic workload with known ground truth and asserts the loop converges: it locks the genuinely-safe downgrade (`after_safe`→cheap), escalates the unsafe exploration candidate (`after_risky`→capable) and the unsafe operator downgrade (`after_routine`→capable), and spends strictly less than a capable-only baseline (~31% cheaper). No live upstream, no randomness — the offline form of the benchmark (E3).

## Deferred (documented)

The verification-gated, **in-request** "resumable cheap attempt" (route cheap → judge the response → re-run on the capable tier if inadequate, so a wrong cheap route costs ~1x not ~2-3x) is intentionally **not** here: it is an invasive change to the execution path that deserves its own design. Trial-based discovery captures most of the cost-discovery value at far lower risk.

## Config (off by default)

```yaml
policy_table:
  adequacy:
    enabled: true          # required for exploration
    explore_enabled: false # opt-in aggressive half
    explore_tier: cheap
    explore_interval: 5    # ~1 in N eligible requests is a trial
    explore_threshold: 3   # adequate trials before locking
```

Validated at parse time: unknown `explore_tier`; `explore_enabled` without `explore_tier`; `explore_enabled` without `enabled`.

## Tests / checks

Ledger (cadence, lock threshold, failed-trial-pins, failed-locked-route-unlocks); router (trial / locked / pin-overrides-exploration / unseen-stays-escalation / guardrail-clamps-a-trial); observer (trial-fail-pins, adequate-trial-locks, cadence-advance); the convergence eval; config defaults + 3 validation cases. `cargo nextest -p bitrouter-sdk -p bitrouter --all-features`: **818 pass**. clippy `-D`, fmt, `generate-schema --check`, rustdoc `-D`, `config validate` all clean.

### DIFF
--- a/apps/bitrouter/src/adequacy/eval.rs
+++ b/apps/bitrouter/src/adequacy/eval.rs
@@ -1,0 +1,183 @@
+//! Convergence eval — the offline form of the adequacy benchmark (E3).
+//!
+//! It drives the *whole* loop — ingress [`PolicyTableRouter`] + the
+//! [`AdequacyObserveHook`] + the [`AdequacyLedger`] — over a synthetic workload
+//! with known ground truth, and asserts the ledger converges to the
+//! cost-optimal-yet-safe policy: it *discovers and locks* the downgrades that are
+//! genuinely safe, and *escalates and keeps off* the ones that are not, with no
+//! round structure and no randomness. This is the test-shaped counterpart of an
+//! offline benchmark; it needs no live upstream because the outcome of each
+//! simulated request is decided by the workload's ground truth.
+
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use bitrouter_sdk::BitrouterError;
+use bitrouter_sdk::caller::CallerContext;
+use bitrouter_sdk::config::{AdequacyConfig, PolicyTableConfig};
+use bitrouter_sdk::language_model::types::{
+    Content, GenerationParams, Message, PipelineRequest, Prompt, ProviderMetadata, Role,
+};
+use bitrouter_sdk::language_model::{ObserveHook, PipelineContext, RequestOutcome};
+
+use crate::adequacy::AdequacyLedger;
+use crate::adequacy::observer::AdequacyObserveHook;
+use crate::policy_table_router::{PolicyTable, PolicyTableRouter};
+
+const CHEAP: &str = "vendor/cheap";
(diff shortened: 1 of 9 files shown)


FrozenArcher404:
Consolidated into #627 — the entire feature (P1 + P2 + P3) in a single branch against `main`. Closing this stacked PR; the commit and its review history are preserved in #627.

PR #678, opened 2026-07-09
feat(providers): add Antigravity BYO subscription

SPIKESPIGEL404:
## What

Routes the user's **Google Antigravity** session (the official `agy` CLI) through bitrouter as a first-party subscription provider, distinct from the `google` API-key path. Stacked on #676 (SuperGrok) — review that first; this PR's diff is Antigravity-only.

```
bitrouter providers login antigravity      # imports the `agy` OS-keyring session
… then route:  antigravity:google/gemini-2.5-flash
```

## How it works

Antigravity diverges from the other BYO-subscription providers in three ways, all handled without a `bitrouter-sdk` core change:

1. **Credential = OS keyring, cross-platform.** `agy` stores its Google OAuth token via Go's `go-keyring`. `import/antigravity.rs` reads it through the `keyring` crate (macOS Keychain / Linux Secret Service / Windows Credential Manager), decodes the `go-keyring-base64` envelope, gates on the `consumer` login, and parses the RFC 3339 expiry.
2. **Confidential refresh, secret not vendored.** Refreshing the Google `1//…` token needs `agy`'s client id **and** a `GOCSPX-…` secret. We pin the *public* client id and read the secret from the **local `agy` binary at refresh time** (`antigravity/agy_client.rs`), trying each candidate — bitrouter ships no Google secret. New `oauth::refresh::refresh_with_client_secret`.
3. **Custom protocol, composed not forked.** cloudcode-pa is Gemini `generateContent` with a different envelope. `antigravity/protocol.rs` registers a `Custom("antigravity")` Adapter/Transport that **composes** the built-in `GenerateContentAdapter`: retargets the URL to `.../v1internal:{verb}` and unwraps the `{"response": …}` envelope on the response **and every SSE chunk**. Registered on the dispatch at app startup. The applier (`antigravity/mod.rs`) adds the Bearer + `antigravity/*` spoof headers, a cached `loadCodeAssist` project bootstrap, and the `{model, project, request}` request envelope.

Plus registry (`antigravity.yaml` + a `RegistryProtocol::Antigravity` variant and its dist-helper mirror), rebuilt dist + docs tables (en/zh), CLI dispatch, and skill.

## Test plan

- `cargo test -p bitrouter-providers --features pkce` → **184 pass** (18 antigravity: import decode/expiry/gate, secret scan, protocol unwrap + endpoint, applier bearer/headers); `dist-helper` 23 pass.
- `clippy` (providers + app + dist-helper) clean; `fmt` clean; providers builds with and without `pkce`.
- Cross-platform keyring **interop confirmed live on macOS** (keyring-rs reads `agy`'s go-keyring item). Linux/Windows implemented via the same crate; Windows target-name scheme is **pending live verification** on a real `agy` login.
- **Live end-to-end against a real `agy` subscription**: `antigravity:google/gemini-2.5-flash` returns completions (non-stream + streaming) — exercising import, confidential refresh via the extracted secret, `loadCodeAssist` project, envelope, and the response/stream `.response` unwrap.

## Risk note

Unofficial: `v1internal` is a private Google surface and the request path presents as the `agy` client (UA + Client-Metadata). It uses the user's own Google account and bitrouter embeds no Google secret, but it can violate Google's ToS — documented as at-your-own-risk in the registry entry and skill.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,17 @@
 version = 4
 
 [[package]]
+name = "aes"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
+dependencies = [
+ "cfg-if",
+ "cipher",
+ "cpufeatures 0.2.17",
+]
+
+[[package]]
 name = "agent-client-protocol"
 version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -195,6 +206,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-broadcast"
+version = "0.7.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "435a87a52755b8f27fcf321ac4f04b2802e337c8c4872923137471ec39c37532"
+dependencies = [
+ "event-listener",
+ "event-listener-strategy",
+ "futures-core",
+ "pin-project-lite",
+]
+
+[[package]]
 name = "async-channel"
 version = "2.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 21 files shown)


SPIKESPIGEL404:
Revised per review: consolidated under the already-staged **`google-ai`** subscription provider instead of adding a separate `antigravity` provider. `registry/providers/google-ai.yaml` is repointed at the real (live-tested) cloudcode-pa Code Assist backend + the `antigravity` custom protocol; `antigravity.yaml` is removed. `PROVIDER_ID = "google-ai"`, login is `bitrouter providers login google-ai`, route is `google-ai:google/gemini-2.5-flash`. Re-verified end-to-end live. The internal Rust module stays named `antigravity` (it's the `agy`/cloudcode-pa mechanism that powers `google-ai`).

PR #712, opened 2026-07-13
chore(registry): trim vendor pool to cap routing at 50

SPIKESPIGEL404:
## Why

This is for the **serving capability-routing** path: the router picks among candidate models, and we want to keep that routing model pool from exceeding **50**. Removing lower-priority vendors keeps the candidate set bounded (now **48 canonical models / 50 providers**).

## What

**Removed model vendors** (`registry/models/`): `baidu`, `bytedance`, `mistralai`, `nvidia`, `kuaishou`, `inclusionai`, `nex-agi`, `meta`.

**Removed providers** left with no models after the vendor cuts:
- `volcengine` — served only ByteDance/Doubao
- `streamlake` — served only Kuaishou / kwaipilot KAT Coder
- `meta-ai` — served only Meta Muse Spark

**Trimmed** the removed vendors' entries out of shared providers: `bitrouter`, `openrouter`, `qianfan`, `qianfan_cn`, `gmicloud`, `novita`, `siliconflow`, and the stale `vertex` scaffold comments.

**Regenerated** `dist/registry/{models,providers}.json` and the `supported-*` docs tables (EN + ZH) via `dist-helper registry build` + `registry docs`. `dist-helper check` passes.

## Notes

- Data-only change (registry YAML + generated dist/docs); no Rust source touched.
- Even open/closed split preserved: **24 open-weight / 24 closed-weight**.


SPIKESPIGEL404:
Open question: whether to re-include SOTA models from NVIDIA (e.g. Nemotron). Depends on accessibility & routing needs — NVIDIA's models are free to serve via NVIDIA AI, so the cost/benefit of keeping them in the routing pool differs from the paid vendors we trimmed here.

PR #743, opened 2026-07-21
feat(substrate): detach/reattach session robustness

SPIKESPIGEL404:
## What & why

Make a manager able to **detach from a running ACP substrate session and reattach later** without losing the turn's outcome, the "a turn is running" signal, or a pending permission. ACP v1 binds turn-completion and permission round-trips to the live connection, so today a manager that detaches mid-turn (or a permission arriving while detached) is lost. This lifts those interactions onto the session's durable, replayable stream.

Full design + rationale (incl. an ACP v1/v2 deep-dive): [`internal/REATTACH_SPEC.md`](internal/REATTACH_SPEC.md).

## Phase 1 — session-scoped permission registry

Fixes: a permission arriving after a manager reattaches was silently never delivered (the upstream stream was **take-once**), and a permission outstanding at detach auto-denied.

- `PendingPermission` is now cloneable with a **shared once-only resolver**; a consumer drop no longer auto-denies while the session lives, so an outstanding permission **survives detach and is re-issued on reattach**.
- New `permissions::PermissionRegistry` is the sole consumer of the upstream stream; `Session::permissions()` is **re-subscribable** (replays outstanding, then streams new).
- Explicit `PendingPermission::deny()` for the "no manager ever" paths (headless `acp prompt`, fleet auto-policy) that previously relied on drop-to-deny.

## Phase 2 — durable, replayable turn state

Fixes: turn completion was invisible on reattach (v1 has no `session/update` for "turn ended"), and there was no "running" signal.

- `turn_state::TurnState` (`running` / `idle{stopReason}` / `requires_action`), shaped field-for-field like ACP **v2 `state_update`**, emitted live **and** to the durable transcript with a per-session `turn_seq`, encoded as a `_bitrouter/turn_state` extension notification (default v1 build).
- **Managers key completion off this stream; the `PromptResponse` is a bare ack** for them — so the future ACP v2 flip is a no-op for consumers.
- Transcript: `Result` → `TurnEnd{turn_seq, stop_reason}`, added `TurnStart`, `Error` gained `turn_seq`; the reader still tolerates legacy `result` lines.
- `down.rs` forwards live turn state and `replay_transcript` re-emits it, so a reattached manager sees turn boundaries + completion.

## Scope / deferred

- **TUI consumer** of `_bitrouter/turn_state` — separate change (by request). GUI is a separate repo.
- **ACP v2 wire encoding** (`acp_v2` feature) — deferred; v2 is still `unstable_protocol_v2`/alpha upstream (spec §0, §7, §10 Phase 3).
- **Turn-state event-ordering fix** — see the follow-up comment below (spec §11.2).

## Verification

- **87** substrate tests, **775** app tests pass (incl. `warm_reattach_socket_e2e`); `cargo clippy` + `cargo fmt --check` clean on both changed crates.
- Reviewed by an adversarial code-review agent; one HIGH finding (fleet auto-policy hang) found and fixed in this PR, one MEDIUM (ordering) deferred with tracking.


SPIKESPIGEL404:
### Follow-up: turn-state event ordering (deferred, spec §11.2)

Surfaced by review. `idle`/`TurnEnd` can be emitted **before** a turn's trailing `session/update`(s), because updates and turn-state reach the transcript (and the live wire) via **two unsynchronized paths** — updates through the `take_transcript_feed` pump, `TurnEnd` directly from the turn worker. On a multi-thread runtime the worker can win the race and write `TurnEnd` before the pump drains the turn's last update (commonly a trailing `usage_update`). On replay, a manager then sees `idle` before that update, weakening the "idle = authoritative completion" contract.

**Why it's not blocking this PR:**
- This PR *exposed* the race rather than introducing it — pre–Phase-2 the `result` line was never replayed, so its ordering was invisible.
- Practical impact today is low: the trailing item is almost always `usage_update` (a running total), not content; and there is **no live consumer yet** (TUI deferred).

**Proper fix** (belongs with the TUI-consumer work, the first live consumer): spec §3's single ordered `SessionEvent` stream — make the update pump the **sole writer** and route turn-state through it with a biased drain-before-`TurnEnd`, so one orderer serves both the transcript and the live forwarder. That's a moderate `engine.rs` event-plumbing refactor; doing it now and reworking it for the TUI pass would be churn.

cc: to be picked up alongside the `_bitrouter/turn_state` TUI consumer.

Issue #787, opened 2026-08-09
feat: decide whether bitrouter/auto should work on a fresh install


SPIKESPIGEL404:
## Context

[#779](https://github.com/bitrouter/bitrouter/pull/779) shipped `@auto` as the public model for the workflow-optimization loop and explicitly rejected `bitrouter/auto` as an alias. We have since reversed that naming decision and shipped `bitrouter/auto` as the canonical public slug (see the linked PR).

The reversal rests on an adoption argument: `vendor/auto` is the convention the gateway ecosystem already uses, so a user pointing an existing config at BitRouter changes **one segment** of a model id rather than learning a new addressing syntax.

That argument is only half-cashed today, and this issue is to decide whether we cash the rest.

## The gap

`openrouter/auto` works the instant you paste it. `bitrouter/auto` currently returns a `400` unless the user already has:

- a `presets.auto` entry in `bitrouter.yaml`, **and**
- a signed `policy-lock.yaml` binding the `auto` policy

`zero_config()` (`crates/bitrouter-providers/src/apply.rs`) defines no presets, so a fresh install has neither. `presets.auto` only appears after `bitrouter optimize setup` or copying `templates/auto-router`.

So the drop-in swap fails at exactly the moment the convention promised it would succeed — for a user who chose us *because* the swap looked free.

## What shipped (option b)

We took the spelling and kept setup required, investing in the error instead. Requesting `bitrouter/auto` with no bound policy now returns:

```
'bitrouter/auto' needs a preset named 'auto' bound to a routing policy; run `bitrouter optimize setup`
```

This is honest and cheap, and the reserved namespace makes it a clean `400` rather than a generic provider `404`. But it leaves one setup step between paste and working, so the adoption win is partial.

## The decision (option a)

Should `bitrouter/auto` resolve to a **built-in default policy** when no lock is present, so a fresh install works on paste?

**For**
- Fully cashes the adoption argument the rename was made on.
- Matches what every other `vendor/auto` does, so the convention stops being a spelling we borrowed and becomes behaviour we honour.
- Removes the only remaining step between "user pastes a model id" and "BitRouter is routing their traffic".

**Against**
- Ships an opinionated default ladder that routes real tokens without operator review. That cuts against the frozen/signed/operator-owned design #779 spent its whole diff defending.
- Raises "which models?" as a product question with cost and availability consequences per user, not a naming one.
[...]

PR #779, opened 2026-08-08
feat: guide @auto workflow optimization

FrozenArcher404:
## Outcome

Make `@auto` the single public model for the workflow-optimization loop and turn the merged #777 control plane into a guided, auditable first-run experience.

This PR deliberately keeps routing generic: the online daemon sees canonical request traces and provider-qualified routes, never benchmark labels, task answers, framework names, or workflow-specific routing keys. Every baseline and candidate tier continues to execute through the private BitRouter daemon. The ACP agent is an independent quality judge only; it is not a tier backend.

## Product decisions

- Public model: `@auto`. `bitrouter/auto` is not introduced as an alias.
- Setup may discover project-owned eval/benchmark entrypoints, but it never invents a task or silently adopts an ordinary unit-test command.
- Exact argv flags remain the deterministic automation contract.
- Existing `@auto` strong/economy tiers are reused; otherwise a TTY asks for provider-qualified routes and headless execution fails before mutation with exact remediation flags.
- Quality profiles remain qualitative search-order preferences. This release requires an agentic pass plus lower normalized showback cost; latency is observe-only.
- First publication from frozen mode requires affirmative TTY consent or `--enable-adaptive` in headless/CI execution.
- Signed progress-guard semantics are not approximated. Setup rejects an active guard before writing any optimization artifact.

### Guided workflow setup

- Discover semantic `package.json` scripts using the repository's package manager.
- Discover bounded project executables from conventional project entrypoint directories.
- Skip generic test helpers, direct symlinks, symlinked search roots, and path escapes.
- Deterministically sort candidates; auto-adopt exactly one and present numbered choices on a TTY.
- Share the discovery path with `bitrouter init --optimize`.
- Preserve `--workflow-command` plus repeated `--workflow-arg` as exact, shell-free argv.

### `@auto` review loop

- Add typed human/JSON reports for setup, run/review, and status.
- Lead review with the two coupled outcomes users care about: agentic quality and normalized cost delta.
- Show latency explicitly as observe-only, plus route key, candidate digest, lineage state, and the exact next command.
- Preserve evaluator/runtime lock details in structured output.

### Safety

- Keep setup and publication separate; setup never enables adaptive mode.
- Add interactive frozen-to-adaptive publication consent while retaining a non-interactive explicit flag.
- Reject active progress guards before any setup mutation.
- Keep workflow discovery provider/model/task neutral.

### Shippable contract

- Update the in-repo CLI reference, implementation contract, and bundled BitRouter skill.
[...]

PR #788, opened 2026-08-09
feat: publish bitrouter/auto as the router slug

SPIKESPIGEL404:
## Outcome

Make `bitrouter/auto` the documented public model for policy-driven routing.

**This reverses a naming decision made deliberately in #779**, which chose `@auto` and stated that "`bitrouter/auto` is not introduced as an alias" — listing it as a correction to the speculative proposal in bitrouter-docs#65. That reversal is the point of this PR, so the reasoning is recorded below rather than left for the next reader to reconstruct.

## Why reverse it

#779's reasoning weighed the **mechanism**: `@` marks a preset — a policy that routes to other providers — rather than a physical model, and `bitrouter` is a real cloud provider with an `api_base`, so `bitrouter/auto` risks implying tokens go to `api.bitrouter.ai`.

What it did not weigh is **adoption**. `vendor/auto` is the convention the gateway ecosystem already uses. A user pointing an existing config at BitRouter changes one segment of a model id instead of learning a new addressing syntax — and users edit config lines, they do not read docs first.

The ambiguity objection is weaker than it looks under that convention: the vendor segment names *the router being addressed*, not the token destination. `openrouter/auto` also bills upstream providers. When a caller talks to the local daemon, BitRouter genuinely is the gateway they are addressing.

What survives is the narrower concern — collision with the `bitrouter` **registry provider entry** — and this PR resolves it by reserving the namespace end to end rather than by avoiding the name.

## Design

**Stage 0, not an ingress transform.** Resolution lands in `resolve_presets`, the one choke point every consumer passes through (daemon pipeline, `routing_preview`, direct `resolve_route_chain` callers such as `validate_routable_model`). An ingress transform would leave the CLI preview disagreeing with the daemon. The Fusion alias (`bitrouter/fusion`) is an ingress transform because it injects *tools*, which routing cannot express; naming is not that kind of problem.

**The whole `bitrouter/` prefix is reserved.** Stage 0 claims it before any provider lookup, and `dist-helper registry validate` now rejects any catalog **or provider-declared** model id underneath — the latter matters because provider ids outside the curated catalog were previously only an advisory. The namespace was verified empty before claiming it.

**`@auto` keeps working.** It was never an auto-specific name — just the generic `@preset[:variant]` form every preset uses. Existing configs are unaffected; the spelling is simply no longer advertised.

## Two silent hazards closed

Both existed because the public spelling carries neither an `@` nor a colon, so the old shape checks missed it:

- `is_explicitly_routed` (`policy_table_router.rs`) did not recognise the namespace, so the legacy inline policy table would have fingerprinted and rewritten a `bitrouter/auto` request instead of letting it reach its signed policy.
- Both tier-target guards (`policy_lock.rs`, two sites) rejected only `@`-prefixed models, so a lock could name `bitrouter/auto` as a tier and sign a **self-referential policy**.

Each has a regression test.

## Errors

The error surface is the actual deliverable here, since setup is still required:

- No bound policy → `'bitrouter/auto' needs a preset named 'auto' bound to a routing policy; run 'bitrouter optimize setup'`
- Unknown slug → `unknown BitRouter model 'bitrouter/xyz'` (a `400`, not a degraded provider `404`)
- `bitrouter:auto` → rejected with a pointer to the slash form; it would otherwise reach Strategy 1 and dispatch to BitRouter Cloud as a model it does not serve
- `bitrouter/fusion` reaching Stage 0 → reports `server_tools.fusion` is unconfigured

## Deliberately not in scope

**`bitrouter/auto` does not work on a fresh install.** `zero_config()` defines no presets, so the slug reports its missing binding rather than falling back to a default route. Whether it should instead work on paste via a built-in default policy is a policy-authority question, not a naming one — bundling it would make this diff much harder to review. Deferred to #787 for team decision.

**The optimizer's harness env keeps `@{preset}`** (`runner.rs`). It is generic over preset name and the reserved namespace only names `auto`, so composing the slug there would `400` every non-`auto` optimization lineage. Commented in place.

## Consumer-visible contract change

The `model` field in the setup/status/review JSON reports changes from `@auto` to `bitrouter/auto`.

## Follow-ups

- **Cloud namespace coordination.** If BitRouter Cloud ever serves a hosted auto-router, `bitrouter/auto` should mean the same thing local and hosted. Worth confirming with the cloud registry owner before that surface exists.
- **bitrouter-docs#65** was rewritten under #779 to say `@auto, not bitrouter/auto`; it needs to flip back.
- `skills/bitrouter/SKILL.md` is 274 lines against the ~200 guideline. It was already 273 before the one-line namespace entry added here; deep detail went into `references/adaptive-routing.md`. Worth a separate trim.

## Verification

- `cargo nextest run --workspace --all-features` — **2690 passed**, 11 pre-existing conditional skips
- `cargo clippy --workspace --all-features --tests -- -D warnings`
- `cargo fmt --all -- --check`
- `cargo run -p dist-helper -- check`
- `cargo run -p dist-helper -- registry validate` — new rule confirmed to fire by temporarily planting a `bitrouter/auto` catalog entry, then restoring
- `git diff --check`


SPIKESPIGEL404:
Docs counterpart updated: [bitrouter-docs#65](https://github.com/bitrouter/bitrouter-docs/pull/65).

Correcting something I noted in the description: **#65 was never actually flipped to `@auto`.** #779 said a stacked docs PR would update its branch, but that never landed — the branch still reads `bitrouter/auto` throughout, so its spelling now matches this PR with no content change needed. I updated its blocker analysis rather than its prose.

That review surfaced a gap worth stating here, because it bounds what this PR delivers. #65's quickstart sends `model: "bitrouter/auto"` to `https://api.bitrouter.ai/v1`. **This PR does not make that work.** The slug is resolved by the local daemon's Stage 0 against a locally-bound preset and signed lock; the hosted endpoint is a separate service that does not implement it. The reserved-namespace rule here also means `bitrouter/auto` can never reach users through the registry catalog — so if Cloud serves it, it has to be built Cloud-side on purpose.

That is the sharp form of the follow-up in the description. Worth a decision from whoever owns the cloud registry before the hosted surface exists, so the slug does not end up meaning two different things depending on which base URL you point at.

Issue #808, opened 2026-08-11
refactor(sdk): move the OTLP exporter into the SDK behind an `otel` feature


SPIKESPIGEL404:
**Spec: `docs/OTEL_SDK_MIGRATION_SPEC.md`** (lands in its own PR; this issue is self-contained without it).

This is a **pure relocation** — no span, attribute, or behaviour changes. Anything that looks like an improvement belongs in a different PR.

## The problem

OTLP export is BitRouter's observability **interop surface**: how a self-hosted user ships router observations to Langfuse, Grafana/Tempo, Honeycomb, Phoenix, or anything else that speaks OTLP. The hosted deployment points at PostHog, but that is one endpoint choice, not the architecture.

Today it lives in a satellite crate a consumer has to discover on their own, then wire by hand through three call sites (`OtelExporter::new`, `router_wrapper()`, `tracing_subscriber_layer()`).

Three facts decide the shape:

1. **The SDK cannot re-export the crate.** `bitrouter-observe` depends on `bitrouter-sdk` (`crates/bitrouter-observe/Cargo.toml:12`), so "keep the crate, re-export for discoverability" is a dependency cycle. There is no middle path.
2. **The crate boundary bought less than it looked.** The stated benefit was quarantining an upstream that "churns minor-to-minor" (`Cargo.toml:19`). But every crate is `version.workspace = true` and releases in lockstep, so an OTel-forced change bumps the workspace anyway. The boundary buys *compile* isolation — which a default-off cargo feature buys identically.
3. **The boundary hid drift.** Until 2026-08-11 `sdk/lib.rs` called observe a "Prometheus exporter" and `sdk/metrics.rs` referenced `bitrouter_observe::PrometheusHook`, a type that no longer existed. In-crate, `cargo check` would have caught the dead seam years earlier.

## What moves

| From | To | Gate |
|---|---|---|
| `crates/bitrouter-observe/src/otel/*` | `crates/bitrouter-sdk/src/otel/*` | `otel` |
| ↳ `http_layer.rs` (needs axum) | same | `otel` **+ `server`** |
| ↳ `auth_client.rs` | same | `otel-http` |
| `crates/bitrouter-observe/src/acp.rs` | `apps/bitrouter/src/observe_acp.rs` | — |
| `crates/bitrouter-observe/` | *deleted* | — |

[...]

PR #809, opened 2026-08-11
refactor(sdk)!: move the OTLP exporter into the SDK

SPIKESPIGEL404:
Implements #808. **This PR now carries the whole migration** — #810 and #811 were squash-merged into their stacked bases, so what lands here is the pre-flight, the relocation, and the public-API guard as a single 43-file change.

`crates/bitrouter-observe` (~4,400 lines) moves into `crates/bitrouter-sdk/src/otel/` behind default-off features, and the crate is deleted.

### ⚠️ Breaking

- `bitrouter-observe` is **deleted**. `bitrouter-cloud` consumes it from crates.io and must migrate in the same release window (3 files, mechanical: add `otel-http` to the `bitrouter-sdk` features, drop the observe dep, rewrite `bitrouter_observe::otel::` → `bitrouter_sdk::otel::`). Nothing is yanked, so any pinned consumer keeps building.
- `tracing_subscriber_layer` returns `impl tracing_subscriber::Layer<S>` instead of the concrete `OpenTelemetryLayer`, and moved module — it is `otel::subscriber`, not `otel::http_layer`.
- The SDK gains three semver-committed **public** foreign types: `tracing_subscriber::{Layer, LookupSpan}` and `tracing_core::Subscriber`. `tower_http` is *not* public.
- ⚠️ `RUST_LOG=bitrouter_observe=debug` no longer matches the HTTP ingress span. Targets are now pinned to `bitrouter::observe::http` / `bitrouter::observe::cardinality` — namespaced as a subsystem rather than a crate, so they survive the relocation. Documented in `docs/CLI.md`.

### Why this is justified

Recorded in the new `docs/OTEL_SDK_MIGRATION_SPEC.md`, which #808 cited but which never existed.

OTLP export is the SDK's own domain model rendered into an open standard — which span is `chat`, what counts as a hop, when settlement closes is BitRouter semantics, not vendor glue, and it must be identical across deployments or "interop surface" means nothing. So: *interop surfaces ship in the SDK behind default-off features; deployment business logic does not.* The `MetricsRenderer` seam is unchanged — the SDK still owns the trait and never the accumulator.

The issue's original arguments (wiring cost, crate-boundary drift, version lockstep) were checked against the code and **do not hold**; see the review comment on #808. The doctrine text in `sdk/lib.rs`, `sdk/metrics.rs` and `docs/DEVELOPMENT.md` is rewritten accordingly — "the SDK ships exactly one observability seam" is false once `OtelObserveHook` is an SDK type.

**`acp.rs` moved into the SDK, not into `apps/bitrouter`** as #808 proposed. That was impossible: `AcpSpanRecorder` stores the exporter's tracer as a field and obtains it via `pub(crate) tracer_clone()`, so crossing a crate boundary would have forced `opentelemetry_sdk::trace::Tracer` into a public SDK signature.

**Feature shape.** `otel = ["otel-http"]` is the public entry point and always builds; a private `__otel-core` carries the transport-agnostic stack and the `compile_error!` guard. Naming the transport-less core `otel` would have made `cargo add bitrouter-sdk --features otel` hard-fail in the foundation crate, where feature unification means one downstream's mistake breaks every other crate in the graph.

**`http_layer` is split.** `otel::subscriber` (gated `otel`) holds the tracing bridge; `otel::http_layer` (gated `otel` + `server`) holds the axum router wrapper. `tower-http` uses a weak dependency feature (`server = […, "tower-http?/trace"]`) for exact feature-intersection semantics, which plain Cargo features cannot express — measured: `server` → no `trace`; `otel-http` → no axum, no `trace`; `otel,server` → both. Zero new crates; `tower-http` was already in the tree via `reqwest`.

### The public-API guard

A `sdk-public-api` job diffs `cargo public-api --simplified` against a committed baseline, with a sentinel step so an empty listing can't pass the negative gates vacuously. Both controls were exercised: flipping `tracer_clone` to `pub` fails it, and so does a re-export — the latter found a real blind spot, since `cargo public-api` renders a bare re-export under its *local* path, hiding the origin crate.

### Frozen

`plugins.bitrouter-observe`, `BITROUTER_OBSERVE_*`, the `/metrics` banner, `io.bitrouter.observe` and the pinned log targets are **unchanged**. `config.plugins` is an unvalidated `HashMap<String, Value>`, so a renamed key would still parse and telemetry would silently stop with `exporter_wired: false` and no error.

### Verification

`cargo fmt --check`, `clippy --workspace --all-features --tests -- -D warnings`, `nextest --workspace --all-features` (**2756 passed / 0 failed / 11 skipped — identical to the base**), `RUSTDOCFLAGS="-D warnings" cargo doc`, and `cargo check -p bitrouter-sdk` under `--no-default-features` / `otel-http` / `otel-grpc` / `otel,server`. `--features __otel-core` alone fails with the transport `compile_error!`, as intended.

`apps/bitrouter/tests/observe_hierarchy.rs` differs by import paths and one comment only — the load-bearing proof the span hierarchy did not change.

### Merged with `main`

`main` moved on while this sat open, so it is merged in (four conflicts):

- **`.github/workflows/ci.yml`** — the ubuntu cold-build diagnostic scaffolding this branch grew (bounded 15-minute build chunks, explicit cache banking, the live PR-comment stream) is **dropped** in favour of #813's sharded ubuntu jobs, now on `main`. That investigation is what #813 concluded; keeping both would leave two competing workarounds for one problem. This branch's own CI work — the rewritten `feature-isolation` job and the new `sdk-public-api` job — is untouched.
- **`apps/bitrouter/src/acp_cli.rs`** — `attach_observability` keeps `main`'s `cost: Option<CostSink>` parameter and this branch's `bitrouter_sdk::otel::OtelExporter` return type.
- **`docs/DEVELOPMENT.md`** — the crate table keeps `main`'s `bitrouter-tui` row; the `bitrouter-observe` row stays deleted.
- **`docs/README.md`** — keeps `main`'s ACP TUI wording and `*_PLAN.md` entry alongside this branch's `OTEL_SDK_MIGRATION_SPEC.md` entry, and drops the `TUI_FIDELITY_MATRIX.md` entry, which #816 deleted.

The ubuntu failure this PR previously carried a footnote about is resolved: **`test (ubuntu-latest)` and all six of its shards are green.** It was environmental — this PR's old head (`52b03e5a`) and the merged #811 head (`b41efdfc`) had byte-identical trees yet opposite ubuntu outcomes — and #813 fixed it on `main`.

All four `sdk-public-api` gates were re-run against the merge result under the pinned `nightly-2026-05-05` / `cargo-public-api 0.52.0`: `public-api-deps.txt` is **unchanged**, so `main`'s SDK work (#816's ACP rebuild, the `responses` adapter growth) added no public dependency.


release-plz.toml
```diff
@@ -15,7 +15,6 @@ changelog_include = [           # Include commits from the SDK and built-in hook
     "bitrouter-sdk",
     "bitrouter-cloud-sdk",
     "bitrouter-guardrails",
-    "bitrouter-observe",
```
FrozenArcher404, line 18: Blocking before squash merge: this deletion is part of removing the published `bitrouter-observe` crate and moving its public API into `bitrouter-sdk`, but the PR title/body still say `fix(observe): pin tracing targets ahead of the SDK move` and “No relocation happens here.” A squash merge will use that metadata, recording a breaking change as a patch fix and preserving a factually false description in history/changelog automation. Please update the PR title to a breaking conventional-commit subject (for example, `refactor(sdk)!: move the OTLP exporter into the SDK`) and rewrite the body for the current 43-file migration before merging.

FrozenArcher404 (dismissed):
The code changes and targeted verification look sound, but the squash metadata currently misrepresents this breaking migration. Please address the inline blocker before merging.

release-plz.toml
SPIKESPIGEL404: You're right, and the cause is worth recording: #810 and #811 were squash-merged into their stacked bases, so this PR's head (`52b03e5a`) now carries the whole 43-file migration rather than the 16-file pre-flight it was opened for. Title is now `refactor(sdk)!: move the OTLP exporter into the SDK` and the body has been rewritten for the full change — breaking notes for the deleted crate, the `tracing_subscriber_layer` signature and module move, the three new public foreign types, and the `RUST_LOG` target change. The `release-plz.toml` deletion is correct and stays: `bitrouter-observe` no longer exists to include in the changelog.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

SPIKESPIGEL404:
## Superseded by #851

**Do not merge this.** #851 is branched from this PR's head and carries every commit in it, so all the work here ships — the span-schema artifact, the OTel-native ingress span, the `sdk-public-api` job and its pins, the `docs/CLI.md` log-target contract, the `main` merge resolution. What changes is one thing: the destination.

This PR moves the OTLP exporter **into** `bitrouter-sdk`. #851 moves it into a standalone `crates/bitrouter-telemetry` instead, and keeps in the SDK only the contract that crate renders — `bitrouter_sdk::observe`, ungated and dependency-free.

### Why, in this PR's own terms

The reversal is not a disagreement with anything measured here. It turns on the axis `OTEL_SDK_MIGRATION_SPEC.md` D1 **conceded rather than answered**:

> Against the standalone crate, SDK placement is measurably *worse* on three axes, and the decision is to accept that, not to deny it.

D1 bought all three — semver, graph position, containment cost — with a single convenience: `bitrouter-cloud` already takes `bitrouter-sdk`, so a standalone crate "would be a second published dependency delivering code the first one could carry." That is one line of `Cargo.toml`. Meanwhile `OTEL_TIERING_SPEC.md` D4 resolved `tracing_subscriber_layer` as *keep*, which made `tracing_subscriber` 0.3 and `tracing_core` 0.1 **permanent** public dependencies of the foundation crate — two 0.x crates where every minor is breaking.

D1 named its own revisit trigger: *"if `tracing-subscriber` 0.4 ships… the standalone crate becomes the better shape again."* That trigger is written backwards — it fires when the bill arrives, not when the debt is taken on. D4 took it on.

The second thing that changed is this PR's own phase 0. Before it, the span schema existed only as ~90 `KeyValue::new` call sites, so the exporter *was* the contract — which is precisely why "identical across deployments" read as an argument for co-location. With the schema a declared, conformance-checked artifact, the contract can stay while the renderer leaves. `otel/mod.rs` in this branch says as much and calls it "a property, not a plan." #851 is the plan.

### What #851 does *not* argue

`OTEL_TIERING_SPEC.md` withdrew its own version of this (phase 3) on measurement, and **those measurements stand**: the crate-count saving has no beneficiary, and the build-cache saving is ~10s inside a build with a ~27s floor and exactly nil on a cold CI cache. #851 reuses neither. Its *cloud question* section reserved reopening for "a positioning decision needing different evidence than this section gathered," and that is the ground #851 stands on.

It is also a different cut from phase 3. Phase 3 split *within* `otel/` — emission stays, export leaves — and hit the eager-tracer-binding hazard documented at length in that spec. #851 moves emission and export together, which is what the module already is, so that hazard never arises and the `tracer_clone` argument retires without being re-litigated.

### The measured result

`crates/bitrouter-sdk/public-api-deps.txt` loses **exactly** `tracing_core` and `tracing_subscriber`, and nothing else, on this branch's own pinned `nightly-2026-05-05` / `cargo-public-api 0.52.0`. That was written down as a prediction *and* an abandon trigger before the work started, and it held. `feature-isolation` correspondingly upgrades from "the OTel stack is absent by default" to "no `opentelemetry*` crate in `bitrouter-sdk --all-features`" — absence at every feature combination.

### Timing is why this happened now

`bitrouter-cloud` still depends on the deleted `bitrouter-observe` and has not migrated — a fact this branch's own `OTEL_TIERING_SPEC.md` D4 records as outranking everything else in it. So cloud performs **one** migration either way, but only while this PR is unmerged. Merging it first and deciding later would have cost two.

Full reasoning, the arguments that are dead, the as-built deviations, and the acceptance results are in `docs/TELEMETRY_CRATE_SPEC.md` on #851. `OTEL_SDK_MIGRATION_SPEC.md` and `OTEL_TIERING_SPEC.md` are kept and marked there rather than deleted — their constraints are still the best record of what is load-bearing, and only their placement conclusions are superseded.

Closing in favour of #851.

SPIKESPIGEL404:
Closed as superseded by #851 — see the analysis above. Reopen if #851 is rejected; its branch descends from this one, so nothing here is lost.

PR #816, opened 2026-08-13
feat(acp): rebuild the ACP agent surface and add the chat TUI

SPIKESPIGEL404:
Deletes the `launch --tui` hosted-terminal experiment and invests the effort in the **ACP agent endpoint** instead, then builds a thin TUI on top of it. The bet (spec §5): improve the agent surface and every ACP client benefits — including clients we don't write — rather than shipping our own widgets.

Executes all four phases of [`docs/ACP_TUI_PLAN.md`](docs/ACP_TUI_PLAN.md). Companion spec: [`docs/ACP_TUI_SPEC.md`](docs/ACP_TUI_SPEC.md).

## Phase 1 — subtraction

- Removes the `--tui` flag, `spawn::exec_hosted`, and the `hosted_env` overlay. `prepare` / `exec_inherited` stay; plain `launch` is unchanged.
- Deletes the PTY host, VT adapter, terminal emulator, and their fixtures.
- Drops four dependencies: `alacritty_terminal`, `portable-pty`, `termwiz`, `wezterm-input-types`.
- Restores full-catalog `launch` support — the four-harness cut was justified partly by hosted mode, which no longer exists.
- Retires `TUI_FIDELITY_MATRIX.md` and the false "all eight catalog harnesses" claim.

## Phase 2 — prerequisites

- **`fix(daemon)`: rebuild the policy table on reload.** `bitrouter reload` returned `{"status":"reloaded"}` while the daemon kept serving the tiers it started with — only `restart` applied a `policy_table:` edit. A success-reporting silent no-op, and a bug on its own merits.
- Captures both stderr streams to one per-session log under `~/.bitrouter/logs/`.
- Structured launch failure instead of a pre-protocol `exit(1)`.

## Phase 3 — the ACP agent surface

- Forwards the five stable v1 `session/update` variants `translate` was swallowing.
- **Emits router-measured cost as `UsageUpdate.cost`** from the metering store — BitRouter sits in the agent seat, so it can report what no upstream harness can.
- **`providers/list` / `providers/set`** over the schema crate's `unstable_llm_providers` types, dispatched as raw JSON-RPC.
- Four wire-level conformance tests against a real `bitrouter acp serve` subprocess.

## Phase 4 — the TUI

- **`crates/bitrouter-tui`** (~1 460 lines): inline viewport, message/tool-call rendering, permission modal, cost line, log tail, provider picker.
- **`bitrouter chat <agent>`**, sharing `RoutingOptions` with `launch` via clap `#[command(flatten)]`.
- Two prerequisites Phase 3 exposed, landed first: session-scoped cost attribution, and a launch-scoped route override so `providers/set` actually reroutes.

## Two decisions were reversed mid-flight, and why

Recorded in [`ACP_TUI_PLAN.md` §C.1](docs/ACP_TUI_PLAN.md) with spec supersession notes, on evidence Phase 3 produced rather than by re-litigation:

- **The TUI is a crate, not in-process.** Spec §9 justified in-process as the way the TUI reads the metering store directly — but Phase 3 put cost on the wire, so that capability no longer requires it. §9's own strongest argument (a module can silently reach anywhere in the crate, which is how the last TUI accreted itself to death) then stands unopposed. The boundary is now a compiler check: `cargo tree -p bitrouter-tui` shows four dependencies and **no edge to the app crate**.
- **The TUI's data comes from the ACP wire, not `snapshot.rs`.** That layer is the `status --watch` shape, nearly all of which §8.3 forbids the TUI from rendering.

## Breaking change

`feat(launch)!: remove --tui hosted mode` — the flag is gone. Plain `bitrouter launch` (inherited terminal) is unaffected and remains the recommended path.

## Notes for review

- **ACP v1 only.** `unstable_protocol_v2` is not enabled anywhere; v2's `PromptResponse` is `{}` with no `stop_reason`, so a v2 gateway is an `engine.rs`/`turn.rs` rewrite. The crate makes adding a v2 *client* later cheap.
- **No credential crosses the `providers/*` wire.** Asserted at two levels, both against *serialized bytes* rather than struct fields, so `_meta` smuggling fails too.
- **Nothing renders a control that lies.** Cost shows `cost unreported` rather than `$0.00` when nothing is reported, and labels `daemon_wide` when attribution is impossible; the provider picker is hidden entirely when the session can't be rerouted.
- **Several commits fix bugs the tests found, not bugs the plan predicted** — most seriously `Session::telemetry()` hands out its receiver once, so no `UsageUpdate` had ever reached a manager; and the TUI did not draw agent replies at all until the transcript was flushed at turn end rather than session end.

## Verification

- `cargo nextest run --all-features` — **2503 passed**, 0 failed
- `cargo clippy --workspace --all-features --tests -- -D warnings` — clean
- `cargo fmt -- --check` — clean
- `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --all-features --no-deps` — clean

The reload fix, the cost-scoping fix, and the transcript-ordering fix are each mutation-checked: disabling the fix fails its test.

Beyond the suite, `chat` was driven against **real `opencode` 1.18.3** over ACP in a tmux TTY — a plain answer, a real tool call with streamed reasoning and status transitions, real telemetry in the session log, and a readable transcript left in scrollback after `Ctrl-D`.

## Known issue, not introduced by the feature work

`test (ubuntu-latest)` hangs until the runner dies. Tracked in #817 with the full investigation: what is established, what was ruled out, and four mitigations that each failed. **All CI workarounds are reverted** — `ci.yml` is byte-identical to main and the `.config/nextest.toml` this branch added is gone, because a workaround that does not work costs the next person the time to disprove it. Everything else is green on all three platforms.


SPIKESPIGEL404:
CI note: `test (ubuntu-latest)` hangs on this branch and I could not fix it. Tracked in #817 with the full investigation — what is established, what was ruled out, and the four mitigations that each failed.

All CI workarounds have been **reverted** (`8f7a3740`): `.github/workflows/ci.yml` is back to main's version and the `.config/nextest.toml` this branch added is gone. A workaround that does not work costs the next person the time to disprove it, so the tree carries none.

Everything else is green — clippy, doc, doctest, fmt, and `test` on macOS and Windows — and the full suite passes locally (2503 tests).

PR #837, opened 2026-08-20
refactor(skills): rewrite /bitrouter as a CLI runbook

SPIKESPIGEL404:
## Why

`skills/bitrouter` was shaped like an install manual rather than something an agent can execute:

- It opened by asking the user **Local or Cloud** — a fork that costs every user for the minority's decision, and one that is anti-instant, since Cloud requires signup and credits while local BYOK with keys already in the environment needs neither.
- There was **no branch for an already-configured machine**. `grep` found zero mentions of already-installed/already-running against 11 mentions of install, so a user with a month-old daemon asking "why is my route falling back?" got an onboarding question.
- **Harness wiring — the actual payoff — sat at section 6 of 7**, behind a reference table, while section 4 was a Python SDK snippet. The activation target is the harness the skill is being read inside.
- It **hand-rolled in prose what `bitrouter init --yes` does natively**, and never mentioned that the wizard is scriptable or that it emits a JSON result envelope. That contract was buried in a table cell in `references/cli.md`.

## What changed

`SKILL.md` is now a six-step activation runbook (198 lines, under the ~200 cap):

1. **Probe** — `--version` / `status` / `providers list`, with a state table routing to the right step.
2. **Install** — only when missing.
3. **Configure** — drives `bitrouter init --yes --use-detected --harness claude --after launch` and parses the JSON envelope (`providers_configured`, `providers_skipped_interactive`, `harnesses_installed`, `after`/`snippet`). Zero-config BYOK is the escape hatch, not the main path.
4. **Credentials needing a human** — exactly what the envelope reports as skipped.
5. **Wire the harness** — `bitrouter launch -a <id>`, with the restart handoff promoted into the main flow.
6. **Verify** — ends on a routed call visible in `bitrouter status --requests`.

Local-or-Cloud is now asked only when step 3 finds no usable credential.

### Scoped to the CLI

Per direction, the skill teaches the CLI only for now:

- Removed the origin-MCP section from `SKILL.md` (`grep -i mcp` returns nothing).
- Deleted `references/mcp-server.md` (280 lines) and `references/agent-plugin.md` (90 lines) — recoverable from history when MCP returns.
- Stripped the plugin sections from the Claude Code and Codex harness references, which pointed at the deleted file.

### Also

- Description trimmed 149 → 71 words.
- `skills/README.md`: the bare `npx skills add bitrouter/bitrouter` form installs **all three** skills (or prompts on an interactive TTY) rather than selecting `/bitrouter`, so every example now passes `--skill`, plus a `--list` line.

## Verification

Installed the local tree through the real skills CLI: `Local path validated` → `Found 3 skills`, frontmatter parses, description renders. Every `references/` path in the table resolves except the two intentional globs (`harness-*.md`, `migrate-from-*.md`, collapsed to fit the line cap). No dangling links to the deleted files anywhere under `skills/`.

No Rust changed, so the cargo checks do not apply.

## Known gaps, not addressed here

- **There is no `bitrouter complete` subcommand.** With MCP out of scope, an `npx skills add` install has no in-session way to make a routed call — verification is `status --requests`, and real use begins after `bitrouter launch`. The skill is now explicit about that ceiling instead of papering over it.
- **`run-bitrouter-benchmark` and `evaluating-bitrouter-routes` still ship in the default sweep.** Unbundling them is a repo-layout decision, not an edit.


SPIKESPIGEL404:
## Benchmark: this branch vs. the skill published on `main`

Each scenario was run twice — once with this branch's `skills/bitrouter`, once with the version a user gets today from `npx skills add bitrouter/bitrouter --skill bitrouter`. The baseline was verified byte-identical to `origin/main`'s `skills/bitrouter` tree (`7738ae22`), not reconstructed.

### Iteration 2 — final (4 scenarios, 27 assertions)

| Metric | Published `main` | This branch |
|---|---|---|
| Pass rate | 79% ± 21% | **100% ± 0%** |
| Time | 185.0s ± 47.7s | **164.9s ± 20.0s** |
| Tokens | 87,385 ± 5,917 | **83,879 ± 11,543** |

| Scenario | This branch | Published |
|---|---|---|
| Fresh setup, wire Claude Code | **8/8** | 4/8 |
| Already-configured machine, routing drift | 7/7 | 7/7 |
| LiteLLM → BitRouter migration | **6/6** | 5/6 |
| Claude Max + cheap supplement | **6/6** | 5/6 |

### Iteration 1 — before the review feedback (3 scenarios, 21 assertions)

| Metric | Published `main` | Branch @ iteration 1 |
|---|---|---|
| Pass rate | 78% ± 25% | 96% ± 7% |
| Time | 162.8s ± 42.0s | 180.5s ± 9.3s |
| Tokens | 80,342 ± 8,978 | 84,375 ± 6,945 |

The cost regression in iteration 1 reversed after the reframe: a shorter decision path beat reading more references.

### What separates them

Every published-version loss is the same failure: **Local-vs-Cloud presented as a deployment fork before anything else**, including on requests that already answered it. Verbatim from its plans:

> "Question 1 (the important one): local daemon, or BitRouter Cloud?" … "Everything below assumes **local**."

> "**Q1 (the big one — this changes everything below): local or cloud?**"

This branch replaces that with a probe, and orders credentials by cost to the user — subscription first, hosted BitRouter as the recommended default provider, then BYOK.

### Defects this surfaced, both present on `main`

- **`launch -a` harness list was stale.** `SKILL.md` and `references/cli.md` claimed support for `hermes`, `openclaw`, `grok`, `agy`; `apps/bitrouter/src/main.rs:406` says those are no longer launch-supported. Found by reading source — **no assertion caught it.**
- `bitrouter status` prints `running: no`, not `stopped` (`main.rs:187`).
- These commands emit JSON by default; neither version said so.
- The published version's plan invented `openai/gpt-5` / `openai/gpt-5-mini`, neither in `dist/registry/models.json`.

### Limitations — read before trusting the numbers

- **n=1 per cell.** The ± figures are variance *across scenarios*, not run-to-run. They say nothing about flakiness.
- **Plan-only.** Agents produced the commands they would run; nothing was installed or executed. Sequencing and factual accuracy are tested; runtime behavior is not.
- **Iteration 1's fresh-setup eval was flawed** — it pre-supplied the probe result, making "probes first" unpassable. Fixed in iteration 2, where the published version *did* probe correctly. Probe-first is not the differentiator; the deployment fork is.
- **The assertion set is saturated** at 100%. It can no longer detect improvement, and it is derived from the skill's own claims, so it is blind to anything both versions get wrong — as the `launch -a` drift demonstrates.
- **It missed a real regression.** Scoring 27/27, this branch had removed the origin MCP `complete` tool — the only lever that works without relaunching a session. The published version offered it. Closed in `4fc15b8d`; no assertion would have caught it.

### Suggested follow-up

The `launch -a` drift shipped on `main` and no test caught it. The lockstep rule in `CLAUDE.md` is currently enforced by discipline alone — a CI check asserting the skill's command claims against the clap definitions would catch the next one.

Issue #847, opened 2026-09-01
fix(acp): cost reports daemon-wide spend in a session-scoped field


SPIKESPIGEL404:
## Summary

When BitRouter cannot attribute a chat session's traffic, it reports **every caller's** spend in `UsageUpdate.cost` and qualifies the scope with a vendor `_meta` key. ACP defines that field as the session's cost, so any client that does not read our private key — which is every client but ours — shows the user a currency figure that is real, precise, and about somebody else's work.

Opening for discussion rather than as a patch, because the fix is a wire-format change and there is a real trade-off in it.

## What ACP specifies

The [session-usage RFD](https://agentclientprotocol.com/rfds/session-usage) is **Completed** (2026-06-05). The schema we build against (`agent-client-protocol-schema` 1.6.0, `src/v1/client.rs:307`) documents the field as:

```rust
/// Cumulative session cost (optional).
pub cost: Option<Cost>,
```

Scope is not left undefined — it is defined *away*. `cost` means this session's spend. There is no field for any other scope, and the RFD does not discuss multi-caller attribution at all.

## What we do

`apps/bitrouter/src/acp_cli.rs:1352`

```rust
let (summary, scope) = match launch_id {
    Some(id) => (
        store.spend_summary_for_launch(id, window).await.ok()?,
        Scope::Session,
    ),
    None => (store.spend_summary(window).await.ok()?, Scope::Wider),
};
```

The `None` branch reads `spend_summary`, which is daemon-wide across every caller, and puts it in `cost`. The scope then rides in `_meta` under `bitrouter/costScope` (`acp_cli.rs:1369`).

`bitrouter chat` reads that key and renders `all callers $X` in yellow, so **our** client is honest. Zed, or any other conforming ACP client, reads `cost.amount` per spec and renders it as the session's.

The `Wider` branch is reached whenever there is no launch token — a `--direct` session, a harness that does not echo the token back, or an own-auth harness whose credential cannot be attributed. That is not a rare path.

## Why this matters here specifically

This is the failure `crates/bitrouter-tui/src/cost.rs:5` was written about:

> A currency figure with no scope is the single most misleading thing an agent's UI can draw. BitRouter's previous status bar rendered the daemon's total and labelled it the session's, so a user with anything else routing saw a number that was real, precise, and about somebody else's work.

[...]

PR #848, opened 2026-09-01
refactor(acp)!: one ACP stack, from the engine to a flat chat loop

SPIKESPIGEL404:
Retires BitRouter's second ACP stack. `chat`, piped `chat`, and `acp prompt` now drive one connection-level controller through one shared client; `engine::Session` and everything that existed to serve it are gone.

The PR opened describing step 0 alone. It now carries the whole arc, so the title and this body are rewritten to match.

## Why

Two ACP stacks shipped simultaneously: `acp serve` ran the controller from #849/#852, while `chat` and `prompt` ran `engine::Session`. **Client capabilities are declared by the client**, so the same harness could be offered a different surface depending on which command launched it. There has to be one.

The controller spec anticipated the migration but not the fragmentation, and left one product decision — what `UsageUpdate.cost` means — to fall out of a refactor. [`docs/ACP_CONTROLLER_AMENDMENT_1.md`](docs/ACP_CONTROLLER_AMENDMENT_1.md) settles both.

## The gate

[`docs/ACP_SAFETY_INVARIANTS.md`](docs/ACP_SAFETY_INVARIANTS.md) enumerates the 14 guarantees the retired code made — each with where it was enforced, what pinned it, who owns it now, and what pins it now. No PR here deleted an enforcing location until its replacement was merged and tested.

It earned its place twice. Deleting the engine would have taken five tests in `up.rs` with it, two of which pin **I1** (an abandoned permission resolves to reject) and **I10** (the wrapper chain's process group dies) — they were ported, not dropped. And the one invariant with *no* controller equivalent was only visible because the document forced the question: under a transparent proxy, `session/request_permission` is an open JSON-RPC request that only an explicit response closes. "Deny by dropping" had to become "deny explicitly, on every path that can abandon a request", and the compiler finds none of them.

## What landed

| Step | Commit |
|---|---|
| Dissolve the `tui` module into the report layer | `3a0a495a`, `eb18bc23`, `9de80346` |
| Retire `down::serve` and the settlement seam | `19bfb79a` |
| Decorate usage updates with attributed cost | `87c907f5` |
| Show how full the context window is | `67356da5` |
| One ACP client; `acp prompt` on the controller | `69c99f6b` |
| Confirm the child is reaped; stop claiming a join | `299fd03c` |
| Piped `chat` on the shared client | `51fb9cc3` |
| Route control on the client; `chat` on the controller | `8153cff9`, `905bd918` |
| Delete the engine | `e0ecd84d` |
| One flat state machine, and six fixes | `19dcd689`…`6cf0f699` |

## Deliberate behaviour changes

Three, all named in their commits:

1. **The transcript no longer freezes while a permission is open.** The point of one select. Previously `answer_permission` awaited stdin alone, so the tool output explaining *what* the agent wanted permission for arrived only after you had decided.
2. **A signal delivered under a modal is acted on immediately.** Same cause — the shutdown arm is finally polled. Latency, not correctness.
3. **`Ctrl-L` redraws instead of denying a permission** (`6cf0f699`, isolated so it can be reverted alone). Both modals treated *any* control chord as cancel, so asking for a redraw denied the agent's request. Nobody chose that; building the key table is what made it visible.

## Things that were claimed and were not true

Recorded because the fixes are in here:

- The NDJSON `session` line advertised a `controller_instance_id` that joins to spend. `prompt` routes with a launch token, so metering persists null for every one of its rows — the value on the wire was a *claimed* header, which §5.5 treats as evidence and not authorization. `launch_id` is the join, and always was. The test that should have caught it asserted `.get(..).is_some()` on a JSON `null`.
- `chat` rendered "the turn's measured cost". It never has on that path.
- Teardown confirmed the harness child was reaped. It ordered the kill in a tail the SDK skips whenever it drops the transport task, so `kill_on_drop` reached `npx` and not the `node` it spawned.

## Reviewing this

The commits are ordered and each passes on its own. `acp prompt` was migrated before `chat` on purpose: it has byte-level NDJSON assertions, and **`chat`'s interactive loop has no integration test at all** — the three exits are covered only by the manual checks in `apps/bitrouter/src/chat/mod.rs`. That asymmetry drove the sequencing throughout, and it is why the state machine's `step()` is a pure function with a 32-cell key table behind it.

2896 tests pass, 11 skipped. Clippy and fmt clean.

Related: #847 (the wire half is fixed here; closing waits on nothing further).


SPIKESPIGEL404:
## Where to spend review effort

59 files is too many to read evenly. Ranked by what would actually hurt if it is wrong:

**1. `905bd918` — `chat` onto the controller.** The riskiest commit in the PR. It changes transport, credential mode, cost source, route surface and teardown at once, on the one surface with **no integration test**. Worth checking specifically:
- `LocalControllerBinding` revokes on every exit path — the failure paths included. `launch_controlled`'s connect-failure branch and the overlay-render failure both revoke before returning.
- The `?` exits inside the loop route through teardown. That is what `d152a5a5`'s `run`/`drive` split is for; before it, a paint failure mid-loop skipped `ControlledSession::shutdown` entirely.

**2. `69c99f6b` — the shared client.** The permission ledger tracks resolvers by `Weak<PermissionResolver>`, not a strong clone, and that choice is load-bearing: a strong clone silently disables deny-by-drop, which is the bug `permissions.rs` documented having introduced with its own registry. `up.rs`'s `dropping_pending_permission_defaults_to_deny` is the end-to-end proof and is unmodified from before the migration.

There is a matching hazard from the other side, worth verifying in `d152a5a5`: because the ledger holds only a `Weak`, a driver that never *removed* answered map entries would disable the same drop arm. `Effect::Resolve` removes rather than looks up, and `Effect::Cancel` clears the map.

**3. `299fd03c` — the reap.** `AgentProcess::connect_to` ordered the kill in a tail the SDK skips whenever it drops the transport task, so nothing waited for it. The confirmation is now handed to the owner via `AgentProcess::reaped`, which lives outside the connection and survives the drop. Note the guarantee is asymmetric and the invariants doc says so: `ControlledSession::shutdown` awaits it directly; `UpstreamConnection` could only wait inside its own thread, and is deleted in `e0ecd84d` anyway.

**4. The two `!` commits break published SDK surface.** `bitrouter-sdk` is `1.0.0-alpha.27`, so breaking is permitted, but `e0ecd84d` removes the ACP `Pipeline`, its three hook traits, `AcpTarget`, `RoutingTable`, `ConfigAcpRoutingTable` and two `pub use` re-exports. I could not check crates.io reverse-dependencies from here — worth a look if anything outside this repo consumes them.

**5. `19bfb79a` and `e0ecd84d` are deletions, and deletions after migration are the safest kind** — the compiler is the test. Skim them.

Suggested order: [`ACP_SAFETY_INVARIANTS.md`](docs/ACP_SAFETY_INVARIANTS.md) first (it is the contract everything else is checked against), then [`ACP_CONTROLLER_AMENDMENT_1.md`](docs/ACP_CONTROLLER_AMENDMENT_1.md) §4 for the cost decisions, then the commits above in that order.

SPIKESPIGEL404:
## What this PR does not pin

Listed so it is a decision rather than a discovery. All of it is recorded in [`ACP_SAFETY_INVARIANTS.md`](docs/ACP_SAFETY_INVARIANTS.md)'s known-unpinned section too.

**`chat`'s interactive loop has no integration test.** This is the big one and it did not change. The three exits — Ctrl-D/Ctrl-C at an idle prompt, panic, and INT/TERM/HUP — are covered only by the manual `stty` checks documented in `apps/bitrouter/src/chat/mod.rs`. What the state machine buys is that the *decisions* are now a pure function with a 32-cell key table behind it (`the_key_table_is_one_table`); what remains untested is everything the **driver** does with an `Effect` — painting, raw mode, the select's cancel-safety, teardown ordering.

An interactive test wants a pty. That is why `acp prompt` was migrated first and used as the parity oracle throughout.

**Delivery of a denial issued at teardown.** Answering a parked permission resolves its oneshot and lets the handler enqueue a response; whether those bytes reach the agent before the transport goes is not guaranteed, because the SDK's outgoing drain is private to it. What *is* guaranteed is the part that matters — nothing resolves to consent, and the child is killed rather than left parked. `AcpClient::shutdown`'s doc says exactly this rather than the stronger thing it used to claim.

**`ControlledSession::shutdown` returning only after the child is reaped.** Implemented; no test drives a real harness through teardown and asserts no survivor. `shutdown_kills_wrapper_chain_process_group` polls for two seconds *afterwards*, so it cannot distinguish a confirmed kill from a race that was usually won. Deliberately left as-is — tightening the tolerance buys a flaky test, not a guarantee.

**`report_turn` / `spawn_tool_spans` emissions** on the prompt and piped-chat paths.

**`acp prompt`'s `usage` line staying the harness's own.** `prompt` installs no cost bridge (only `serve` and `chat` issue a controller binding), so it never decorates. That follows structurally, and `usage_capability_is_absent_without_a_cost_bridge` covers the mechanism at the controller level — but nothing asserts the absence end to end in the NDJSON.

**One assertion was removed**, and it is the only coverage this PR deletes: that a binding-less controller writes `routeControl: null` rather than omitting the key. It was being asserted from the wrong side (the client, about the controller), the parser treats null and absent identically, and `route_control_capability_needs_all_three_conditions` pins both. Re-homing it properly meant ~40 lines of raw-manager plumbing for a distinction nothing depends on.

PR #877, opened 2026-09-05
feat(cli)!: rename the invocation to `bro`

SPIKESPIGEL404:
`bitrouter <cmd>` becomes `bro <cmd>`.

## Scope: invocation only

The product is still BitRouter. Deliberately **not** renamed:

`bitrouter.yaml` · `~/.bitrouter/` · all 29 `BITROUTER_*` env vars · the `brk_` key prefix · the `io.bitrouter.observe` OTel scope and `bitrouter` meter name · crate / npm / Homebrew-tap names · `AxoUpdater::new_for("bitrouter")` (the dist install-receipt app name — changing it breaks self-update)

Those are wire and config contract. `docs/CLI.md` already commits in writing that the telemetry identifiers are never renamed.

## The alias

`[[bin]]` is now `bro`; `[dist.bin-aliases]` has the shell, PowerShell, npm, and Homebrew installers create a `bitrouter` symlink onto it. Verified against real generated installers (`dist build --artifacts=global`, cargo-dist 0.31.0) rather than assumed — cargo-dist does not reject unknown config keys, so inspecting the output is the only proof.

The alias is **installer-only**; release archives ship `bro` alone. So `cargo install` and raw-tarball users get `bro` only.

**It is removed in 1.0.0.** Reaching the binary through it prints one line to stderr:

```
note: `bitrouter` is now `bro`. This alias is removed in 1.0.0.
```

stderr, never stdout — `bro <cmd> 2>/dev/null | jq .` still yields exactly one clean JSON value, as documented.

## Invoked-name plumbing

New `bitrouter_sdk::invocation` records the `argv[0]` basename once at startup. `parse_cli` sets `.name()`/`.bin_name()` on the built `clap::Command` before `get_matches`, which propagates the typed name into every subcommand's usage line.

It is deliberately **not** lazy: the SDK is also an assembly library, and a lazy read would make an embedder's binary print ``run `their-app cloud login` ``. Unrecorded callers get the canonical `bro`, which is always correct advice — and is what keeps hint assertions in unit tests deterministic.

```
$ bro --help          →  Usage: bro [OPTIONS] [COMMAND]        (stderr: 0 bytes)
$ bitrouter --help    →  Usage: bitrouter [OPTIONS] [COMMAND]  (stderr: the note)

$ bro agents install no-such-agent        →  Run `bro agents list --remote` …
$ bitrouter agents install no-such-agent  →  Run `bitrouter agents list --remote` …
```

## Also in here

- **CI would have broken:** the test matrix had `flags: --bin bitrouter`.
- **Lockstep** (per `CLAUDE.md`): plugin manifests' MCP `command` → `bro`; `skills/bitrouter/` invocations → `bro`. Directory names, skill names, and server keys unchanged.
- **Docs are `bro`-only.** The alias is documented solely as deprecated, not as a second supported spelling.
- **Comments:** every comment naming a subcommand was renamed via an allowlist of the 28 real subcommand names — so prose survived correctly (``a bitrouter run``, ``the bitrouter home``, ``bitrouter consumes``, ``bitrouter server-tool``).
- **The config JSON schema is regenerated**, since `schemars` derives its descriptions from doc comments. `cargo run -p dist-helper -- check` catches this.

## Verification

```
cargo nextest run --all-features   2995 passed, 11 skipped
cargo clippy --all-features        no errors
cargo fmt -- --check               clean
cargo run -p dist-helper -- check  schema + registry up to date
```

Plus manual dual-invocation via a `bitrouter`-named symlink: usage lines, stderr routing, hint routing, `NO_COLOR` gating, and the clean-stdout-JSON contract under both names.

## Known, not addressed

`skills/bitrouter/SKILL.md` is 203 lines, just over the ~200 guideline in `CLAUDE.md`. It was 202 before this change, so the rename did not cause it.


FrozenArcher404:
bro?? Seriously?

FrozenArcher404:
> bro?? Seriously?

Why not just `br`

SPIKESPIGEL404:
> > bro?? Seriously?
> 
> Why not just `br`

more fun

